PR #10, opened 2021-12-27
[13.0][FIX] mrp_unbuild_valuation: price unit assignment improvement

dalonsod:
This fix makes a better produce move unit price assigment, so lets Odoo process to be used instead of making our own further adjustments on stock valuation layers. This approach also lets Odoo AVCO update in Odoo way.


dalonsod:
cc @ChristianSantamaria

PR #12, opened 2022-03-08
[13.0][FIX] mrp_unbuild_advanced_svl: safety deletion of related account moves

dalonsod:
With this fix related unbuild account moves are deleted, in a safe mode. Previous to this fix, account moves were not deleted and, then, unlinked to the original document


dalonsod:
@lmiguens-solvos already installed in Test, could you check?

For older account moves, maybe a little python script should be needed when if we want to emulate the following code: https://github.com/solvosci/slv-manufacture/blob/13.0-fix-mrp_unbuild_advanced_svl-account_move/mrp_unbuild_advanced_svl/models/stock_move.py#L53

lmiguens-solvos:
> @lmiguens-solvos already installed in Test, could you check?
> 
> For older account moves, maybe a little python script should be needed when if we want to emulate the following code: https://github.com/solvosci/slv-manufacture/blob/13.0-fix-mrp_unbuild_advanced_svl-account_move/mrp_unbuild_advanced_svl/models/stock_move.py#L53

LGTM.
Tested the 3 deleted: stock.valuation.layer, account.move and account.move.line

PR #13, opened 2022-07-12
[13.0][ADD] stock_valuation_mrp

dalonsod:
MRP adaptations for brand new stock valuation method

Depends on solvosci/slv-stock#41


dalonsod:
This addon can be also merged because solvosci/slv-stock#41 is already merged. Let's go...

cc @ChristianSantamaria

PR #18, opened 2022-12-09
[11.0][ADD] mrp_project_analytic

dalonsod:
For those MRPs linked to projects, when account moves are generated, project analytic account is automatically set.


dalonsod:
Needed in v15, we'll merge it first, let's go!

PR #19, opened 2023-01-03
[14.0][FIX] fcd_weight_scale_mrp: singleton in default production secondary…

ChristianSantamaria:
… uom


dalonsod:
This is a temporary workaround, proper way will be prevent more than one secondary unit with factor 0.0 in general

PR #21, opened 2023-01-20
[14.0][IMP] fcd_weight_scale_mrp: some improvements

ChristianSantamaria:
Some improvements: 
 - Adds button to reload in scale weight
 - Add checkpoint user to login in res.conf
 - Add autologging on scale with checkpoint user
 - Add Ip filter to weight scale


dalonsod:
🤞🏽 again

PR #27, opened 2023-03-24
[15.0][MIG] mrp_project

dalonsod:
Standard migration


dalonsod:
Tested in a empty database

dalonsod:
cc @adriresu due to Crt inminent migration to v15

PR #28, opened 2023-03-24
[15.0][MIG] product_bom_std_price

dalonsod:
Standard migration


dalonsod:
Tested in an empty database 👍

PR #29, opened 2023-05-31
[13.0][ADD] mrp_unbuild_export_xlsx

ChristianSantamaria:


dalonsod (changes requested):
Please check `static/src` folder (seems to be unnecessary), addons dependency (quality control), missing `oca_dependencies.txt` for `reporting-engine` entry and license (should be AGPL).

PR #38, opened 2023-09-26
[16.0][MIG] mdc

dalonsod:
~~This is a work in progress.~~

This is a basic migration, most validated for installation issues, backend views in general management, and WIN/WOUT and other frontend functionalities.

Not tested for these cases:
* Advanced worksheet functionalities (blind import, worksheet recalculation depending on lot dates,...).
* Reporting in general.


dalonsod:
cc @adriresu

PR #43, opened 2023-10-13
[15.0][ADD] mrp_production_skip_sched

dalonsod:
Prevents that scheduler is triggered when a production is confirmed or changed once confirmed.
Scheduler cron or manual execution should do the trick instead.

This could be achieved setting `stock.no_auto_scheduler` config parameter, but it should affect to other processes, like pickings.


dalonsod:
Tested by customer :+1:

PR #44, opened 2023-10-16
[14.0][FIX] mrp_timesheet_link: secure ``cancel()`` access

dalonsod:
A cancellation by a user without project access should raise a permissions error. This fix it


dalonsod:
cc @ChristianSantamaria

PR #53, opened 2024-02-07
[13.0][ADD] mrp_unbuild_cust_alu_analytics

adriresu:


adriresu:
TODO: 
The unlink method should trigger when deleting a record from bom_quants_ids not when deleting from bom_quants_total_ids.
Currently commented out.

dalonsod:
Related to last change, has frontend (incidence creation and editing) behavior changed?

mrp_unbuild_cust_alu_analytics/__manifest__.py
```diff
@@ -0,0 +1,26 @@
+# © 2024 Solvos Consultoría Informática (<http://www.solvos.es>)
+# License LGPL-3.0 (https://www.gnu.org/licenses/lgpl-3.0.html)
+{
+    "name": "Mrp Unbuild Cust Alu Analytics",
+    "summary": """
+        Add incidences and process type to Unbuild, add custom report, add analytics to unbuilds
+    """,
+    "author": "Solvos",
+    "license": "LGPL-3",
+    "version": "13.0.1.0.0",
```
dalonsod, line 10: Use AGPL-3

mrp_unbuild_cust_alu_analytics/data/ir_cron.xml
```diff
@@ -0,0 +1,12 @@
+<odoo>
+    <record id="cron" model="ir.cron">
```
dalonsod: Cron record should always be created under a `<data noupdate="1">` section

mrp_unbuild_cust_alu_analytics/data/qc_test.xml
```diff
@@ -0,0 +1,234 @@
...
+        <record model="qc.test.question" id="test_line_25">
+            <field name="test" ref="qc_test_1"/>
+            <field name="name">Al</field>
+            <field name="type">quantitative</field>
+            <field name="min_value">0</field>
+            <field name="max_value">100</field>
+            <field name="uom_id" ref="product_uom_percentage"></field>
+        </record>
+    </data>
+</odoo>
```
dalonsod: Missing EOF empty line

mrp_unbuild_cust_alu_analytics/i18n/mrp_unbuild_workflow.pot
```diff
@@ -0,0 +1,506 @@
+# Translation of Odoo Server.
```
dalonsod: This file should be renamed

mrp_unbuild_cust_alu_analytics/data/qc_test.xml
```diff
@@ -0,0 +1,234 @@
...
+        <!-- UOM -->
+        <record model="uom.uom" id="product_uom_percentage">
+            <field name="name">%</field>
+            <field name="category_id" ref="product_uom_categ_percentage"></field>
+            <field name="factor" eval="1.0"/>
+            <field name="rounding" eval="0.001"/>
+        </record>
+
+        <!-- TEST -->
+        <record model="qc.test" id="qc_test_1">
```
dalonsod, line 19: This record is used in code, so you must prevent its deletion. Please protect it in unlink method

mrp_unbuild_cust_alu_analytics/models/mrp_bom.py
```diff
@@ -0,0 +1,9 @@
+# © 2024 Solvos Consultoría Informática (<http://www.solvos.es>)
+# License LGPL-3.0 (https://www.gnu.org/licenses/lgpl-3.0.html)
+
+from odoo import api, fields, models
+
+class MrpBom(models.Model):
+    _inherit = "mrp.bom"
+
+    process_type_id = fields.Many2one('mrp.unbuild.process.type', string='Process type')
```
dalonsod: Missing EOF empty libe

dalonsod: Unnecessary string (Odoo will name "Process Type" this field). Don't change anything, anyway, only take it in account the next case

mrp_unbuild_cust_alu_analytics/models/mrp_unbuild.py
```diff
@@ -0,0 +1,187 @@
...
+class MrpUnbuild(models.Model):
+    _inherit = "mrp.unbuild"
+
+    process_type_id = fields.Many2one('mrp.unbuild.process.type', string='Process type')
+    incidence_ids = fields.One2many(
+        comodel_name="mrp.unbuild.incidence",
+        inverse_name="unbuild_id",
+        string="Incidences",
+    )
+    analytic_ids = fields.Many2many(
```
dalonsod: Why this field is `Many2many`? And why computed? Inspections have an `unbuild_id` field, so it this could simply be defined as a `One2many` field, what do you think?

mrp_unbuild_cust_alu_analytics/models/mrp_unbuild.py
```diff
@@ -0,0 +1,187 @@
...
+
+    def compute_move_analytics(self):
+        for record in self:
+            analytics = self.env["qc.inspection"].search([("unbuild_id", "=", record.id)])
+            record.analytic_ids = analytics
+
+    @api.onchange('product_id')
+    def _onchange_product_id(self):
+        if self.product_id and self.process_type_id:
+            self.bom_id = self.product_id.product_tmpl_id.bom_ids.filtered(lambda r: r.process_type_id == self.process_type_id)
```
dalonsod, line 29: Is it granted that this line will always return a single record (or none)?

mrp_unbuild_cust_alu_analytics/models/mrp_unbuild.py
```diff
@@ -0,0 +1,187 @@
...
+    def _onchange_bom_id(self):
+        if self.bom_id:
+            self.product_uom_id = self.bom_id.product_uom_id.id
+        else:
+            self.product_uom_id = False
+
+    @api.constrains('bom_quants_ids')
+    def _contrains_bom_quants_ids(self):
+        total_product_qty = sum(self.bom_quants_total_ids.mapped('total_qty'))
+        self.product_qty = total_product_qty if total_product_qty > 0 else 1
```
dalonsod: `@api.constrains` shouldn't be used to update values, only for validation purposes.
But this is the question: why this line? Should it be linked to this? 

https://github.com/solvosci/slv-manufacture/tree/13.0/mrp_unbuild_force_exact_qty

If you test your code without that addon installed, this method will become unneeded IMO... If I'm right, please tell me, so the correct approach should be this one:
- Improve of `mrp_unbuild_force_exact_qty` that enable a bypass confirming unbuild.
- Then, adding that bypass here, in this addon.

cc @ChristianSantamaria

mrp_unbuild_cust_alu_analytics/models/mrp_unbuild.py
```diff
@@ -0,0 +1,187 @@
...
+
+    @api.constrains("incidence_ids")
+    def _constrains_incidences_ids(self):
+        if self.incidence_ids:
+            self.shift_stop_time = sum(self.incidence_ids.mapped('duration'))
+        else:
+            self.shift_stop_time = 0
```
dalonsod, line 45: the same as above: `@api.constrains` is not the proper way for this. Could you try it in other way?

mrp_unbuild_cust_alu_analytics/models/mrp_unbuild.py
```diff
@@ -0,0 +1,187 @@
...
+                "&",
+                ("process_type_id", "=", record.process_type_id.id),
+                ("product_tmpl_id", "=", record.product_id.product_tmpl_id.id)
+            ]
+            domain_product_id = [
+                ("process_type_ids", "=", record.process_type_id.id),
+            ]
+        return {"domain": {"bom_id": domain_bom_id, "product_id": domain_product_id}}
+
+    def read_qc_inspection_xml(self):
```
dalonsod: Better name (this is a protected method, only called by a cron):
```suggestion
    def _read_qc_inspection_xml(self):
```

mrp_unbuild_cust_alu_analytics/models/mrp_unbuild.py
```diff
@@ -0,0 +1,187 @@
...
+        # Find the folder and create if not exists
+        path = self.env.company.xml_analytic_path
```
dalonsod: This only be executed for active company for cron user (usually OdooBot). Don't change anything, but add a TODO comment, in order to make this real multi-company

mrp_unbuild_cust_alu_analytics/models/mrp_unbuild.py
```diff
@@ -0,0 +1,187 @@
...
+        logger = logging.getLogger(__name__)
+        if not path:
+            logger.critical(
+                "XML Analytic Path not defined"
+            )
+            raise UserError(_("XML Analytic Path not defined"))
```
dalonsod: Why not create error message once? More examples below

mrp_unbuild_cust_alu_analytics/models/mrp_unbuild.py
```diff
@@ -0,0 +1,187 @@
...
+                                        analytic['attributes']['ElementName']: float(element['children'][0]['text'])
+                                    })
+
+                    #Process xml data
+                    if unbuild_id:
+                        if unbuild_id.state == 'done':
+                            move_id = False
+                            move_line_ids = unbuild_id.produce_line_ids.move_line_ids
+                            for move_line in move_line_ids.move_id:
+                                if "Virtual Locations" in move_line.location_dest_id.complete_name:
```
dalonsod: This is a bad approximation... you mean "the move that has Virtual Locations/Production as destination location", don't you? If I'm right, this is strictly better and not depending on translations, etc:
```suggestion
                                if move_line.location_dest_id.usage == "production":
```

mrp_unbuild_cust_alu_analytics/models/mrp_unbuild_bom_quants.py
```diff
@@ -0,0 +1,17 @@
...
+# License LGPL-3.0 (https://www.gnu.org/licenses/lgpl-3.0.html)
+
+from odoo import api, fields, models, _
+
+class MrpUnbuildBomQuants(models.Model):
+    _inherit = "mrp.unbuild.bom.quants"
+    
+    def save_and_new(self):
+        self.ensure_one()
+        self.unbuild_id.product_qty = sum(self.unbuild_id.bom_quants_total_ids.mapped('total_qty'))
```
dalonsod: Related to above comment of "totally exact quantity" issue?

mrp_unbuild_cust_alu_analytics/models/mrp_unbuild_bom_quants.py
```diff
@@ -0,0 +1,17 @@
...
+        return super(MrpUnbuildBomQuants, self).save_and_new()
+
+    def save_and_close(self):
+        self.ensure_one()
+        self.unbuild_id.product_qty = sum(self.unbuild_id.bom_quants_total_ids.mapped('total_qty'))
```
dalonsod: Related to above comment of "totally exact quantity" issue?

mrp_unbuild_cust_alu_analytics/models/mrp_unbuild_bom_total.py
```diff
@@ -0,0 +1,18 @@
...
+class MrpBom(models.Model):
+    _inherit = "mrp.unbuild.bom.totals"
+
+    total_qty_percentage = fields.Float(
+        widget="percent",
+        compute="_compute_total_qty_percentage",
+        store=True
+    )
+
+    @api.depends("expected_qty")
```
dalonsod: `@api.depends` for this compute method is very strange, because `expected_qty` isn't used within method, and other data that is actually used is not added to depends clause. Is it right?

mrp_unbuild_cust_alu_analytics/security/ir.model.access.csv
```diff
@@ -0,0 +1,3 @@
+id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
+access_mrp_unbuild_process_type,access_mrp_unbuild_process_type,model_mrp_unbuild_process_type,base.group_user,1,1,1,1
+access_mrp_unbuild_incidence,access_mrp_unbuild_incidence,model_mrp_unbuild_incidence,base.group_user,1,1,1,1
```
dalonsod: Missing EOF empty line

mrp_unbuild_cust_alu_analytics/security/ir.model.access.csv
```diff
@@ -0,0 +1,3 @@
+id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
+access_mrp_unbuild_process_type,access_mrp_unbuild_process_type,model_mrp_unbuild_process_type,base.group_user,1,1,1,1
```
dalonsod: Should any user create and delete process types? cc @ChristianSantamaria

mrp_unbuild_cust_alu_analytics/views/mrp_bom_views.xml
```diff
@@ -0,0 +1,27 @@
...
+        </record>
+
+        <!-- Tree View -->
+        <record id="mrp_bom_form_process_type_tree_view" model="ir.ui.view">
+            <field name="name">mrp.bom.view.tree (process_id)</field>
+            <field name="model">mrp.bom</field>
+            <field name="inherit_id" ref="mrp.mrp_bom_tree_view"/>
+            <field name="arch" type="xml">
+                <xpath expr="//field[@name='type']" position="before">
+                    <field name="process_type_id" optional="0"/>
```
dalonsod: Optional, Show or Hide?

mrp_unbuild_cust_alu_analytics/views/mrp_bom_views.xml
```diff
@@ -0,0 +1,27 @@
...
+                </xpath>
+            </field>
+        </record>
+    </data>
+</odoo>
```
dalonsod: Missing EOF empty line

mrp_unbuild_cust_alu_analytics/views/mrp_unbuild_bom_total_views.xml
```diff
@@ -0,0 +1,14 @@
...
+            <field name="model">mrp.unbuild.bom.totals</field>
+            <field name="inherit_id" ref="mrp_unbuild_bom_cust_qty.mrp_unbuild_bom_total_view_tree"/>
+            <field name="arch" type="xml">
+                <xpath expr="//field[@name='expected_qty']" position="after">
+                    <field name="total_qty_percentage" widget="percentage"/>
+                </xpath>
+            </field>
+        </record>
+    </data>
+</odoo>
```
dalonsod: Missing EOF empty line

mrp_unbuild_cust_alu_analytics/views/mrp_unbuild_process_type_views.xml
```diff
@@ -0,0 +1,48 @@
...
+    <record id="action_mrp_process" model="ir.actions.act_window">
+        <field name="name">MRP Process</field>
+        <field name="res_model">mrp.unbuild.process.type</field>
+        <field name="view_mode">tree,form</field>
+    </record>
+
+    <!-- Menu Item -->
+    <menuitem id="menu_mrp_process" name="Process Types" parent="mrp.menu_mrp_configuration" action="action_mrp_process"/>
+
+</odoo>
```
dalonsod: Missing EOF empty line

mrp_unbuild_cust_alu_analytics/views/mrp_unbuild_views.xml
```diff
@@ -0,0 +1,95 @@
...
+        </record>
+
+        <!-- Tree View -->
+        <record id="mrp_unbuild_form_process_type_tree_view" model="ir.ui.view">
+            <field name="name">mrp.unbuild.view.tree (process_id)</field>
+            <field name="model">mrp.unbuild</field>
+            <field name="inherit_id" ref="mrp.mrp_unbuild_tree_view"/>
+            <field name="arch" type="xml">
+                <xpath expr="//field[@name='product_id']" position="before">
+                    <field name="process_type_id" optional="0"/>
```
dalonsod: Show or Hide?

mrp_unbuild_cust_alu_analytics/views/mrp_unbuild_views.xml
```diff
@@ -0,0 +1,95 @@
...
+            </field>
+        </record>
+
+        <record id="mrp_unbuild_form_group_shift_mrp_unbuild_workflow" model="ir.ui.view">
+            <field name="name">mrp.unbuild.form (group_shift) (mrp_unbuild_workflow)</field>
+            <field name="model">mrp.unbuild</field>
+            <field name="inherit_id" ref="mrp_unbuild_advanced.mrp_unbuild_form_view_user"/>
+            <field name="arch" type="xml">
+                <xpath expr="//notebook" position="before">
+                    <group
```
dalonsod: Why these changes? If needed, we should consider making them as PR improvement in original module.... cc @ChristianSantamaria

mrp_unbuild_cust_alu_analytics/views/mrp_unbuild_views.xml
```diff
@@ -0,0 +1,95 @@
...
+                    </page>
+                    <page string="Analytics">
+                        <field name="analytic_ids" class="oe_editable_tree_only">
+                        </field>
+                    </page>
+                </xpath>
+            </field>
+        </record>
+    </data>
+</odoo>
```
dalonsod: Missing EOF empty line

mrp_unbuild_cust_alu_analytics/views/res_config_settings_view.xml
```diff
@@ -0,0 +1,29 @@
...
+                                </div>
+                            </div>
+                        </div>
+                    </div>
+                </div>
+            </xpath>
+        </field>
+    </record>
+
+</odoo>
```
dalonsod: Missing EOF empty line

requirements.txt
```diff
@@ -0,0 +1,2 @@
+lxml==4.3.2
+openpyxl==3.1.2
```
dalonsod: Missing EOF empty line

requirements.txt
```diff
@@ -0,0 +1,2 @@
+lxml==4.3.2
```
dalonsod: lxml is already an Odoo requirement (see https://github.com/odoo/odoo/blob/4cc086d330b73514fbc64a7fcf22d8a7a9f1b691/requirements.txt#L14). Why this?

dalonsod (changes requested):
See comments, only initial code review

adriresu:
> Related to last change, has frontend (incidence creation and editing) behavior changed?

Finally any change was required, and It works like before, however, all the incidences created before this commit going to continue existing in DB but inaccessible from odoo views.

dalonsod:
That's right, don't worry, because we're already in testing step. But when we'll update in AluDev Environment please report other users, because example data for this should be re-created.

mrp_unbuild_cust_alu_analytics/__manifest__.py
adriresu: Done

mrp_unbuild_cust_alu_analytics/data/ir_cron.xml
adriresu: Done

mrp_unbuild_cust_alu_analytics/data/qc_test.xml
adriresu: Done

mrp_unbuild_cust_alu_analytics/i18n/mrp_unbuild_workflow.pot
adriresu: Done, but i need to recreate translations, too many changes from last translation

mrp_unbuild_cust_alu_analytics/views/mrp_unbuild_views.xml
adriresu: Done

mrp_unbuild_cust_alu_analytics/views/res_config_settings_view.xml
adriresu: Done

requirements.txt
adriresu: Done

requirements.txt
adriresu: mb, I didn't see that it was on requirenments, otherwise it is specified to python 3.7 versions or older, locally i'm using python 3.8

dalonsod: nope, Odoo points python 3.7 and newer, check it

mrp_unbuild_cust_alu_analytics/models/mrp_unbuild.py
dalonsod: Still pending to answered/fixed

mrp_unbuild_cust_alu_analytics/models/mrp_unbuild.py
dalonsod: Pending to be answered

mrp_unbuild_cust_alu_analytics/models/mrp_unbuild.py
dalonsod: Pending to be answered/fixed

mrp_unbuild_cust_alu_analytics/models/mrp_unbuild.py
dalonsod: Pending to be answered/fixed

mrp_unbuild_cust_alu_analytics/models/mrp_unbuild.py
dalonsod: Pending to be answered/fixed

mrp_unbuild_cust_alu_analytics/models/mrp_unbuild.py
dalonsod: Pending to be answered

mrp_unbuild_cust_alu_analytics/models/mrp_unbuild.py
dalonsod: Pending to be answered/fixed

mrp_unbuild_cust_alu_analytics/models/mrp_unbuild.py
dalonsod: Pending to be answered/fixed

requirements.txt
adriresu: Yes, that's true, sorry

mrp_unbuild_cust_alu_analytics/data/qc_test.xml
adriresu: Done

adriresu:
At the last commit, datetime is converted based on user timezone, but timezone should be removed before printing in Excel.
Do u know some better practice to get this? @dalonsod

dalonsod:
Check this:
https://github.com/odoo/odoo/blob/4cc086d330b73514fbc64a7fcf22d8a7a9f1b691/odoo/fields.py#L1812-L1837

That method is used along some Odoo addons

mrp_unbuild_cust_alu_analytics/models/mrp_unbuild.py
adriresu: By the user's side, I think so, but I can't assure.
Should it be protected anyway?

mrp_unbuild_cust_alu_analytics/views/mrp_unbuild_views.xml
adriresu: Solved

mrp_unbuild_cust_alu_analytics/models/mrp_bom.py
adriresu: Solved

mrp_unbuild_cust_alu_analytics/models/mrp_unbuild.py
adriresu: Solved

mrp_unbuild_cust_alu_analytics/models/mrp_unbuild.py
adriresu: Remmoved logger, with UserError is enought

mrp_unbuild_cust_alu_analytics/models/mrp_unbuild.py
adriresu: Solved

mrp_unbuild_cust_alu_analytics/security/ir.model.access.csv
adriresu: Solved

ChristianSantamaria:
@adriresu @dalonsod 
I have added some improvements from today's meeting to the addon, I have tested it and it works correctly but it is not a stable version since I have done it running so as not to crash with AR. This addon the following:
- New menus to view incidents
- New incident type model
- Some translations
- Delete some fields from the incidents that were not wanted
- And change the name of the field from Stage to State

There are still some things to do, I tried but that generation of xlsx is difficult to follow... I would need two things:
- Add the name of the Unbuild in the second column
- Add between "Rest Break Time" and "Gross Time" a new column for a new field (Which I haven't done yet...) that will be "Scheduled Time" I will create this field in the other addon...

If you have any questions, tell me

adriresu:
In my previous commit, I split the display_name with string split to get only the name without the reference_code, However this solution has a limitation, which is when the name itself has brackets (I splitted the display_name based on brackets)
I considered regex to solve this, but i don't know if there is some built-in odoo solution to fix this.
@dalonsod

dalonsod:
Have you tried simply get `name` from product, instead of `display_name`?

imagen

adriresu:
> Have you tried simply get `name` from product, instead of `display_name`?
> 
> imagen

it could work, but for subproducts, name dont has the subproduct tags:
image

dalonsod:
OK, I know. Please add a `TODO` to code with this, and keep current workaround. I'm sure I had the same problem in other development, but I don't remember where, so I can't add an alternative solution right now.

dalonsod:
I've finally found it!!

imagen

adriresu:
> I've finally found it!!
> 
> imagen

I'm going to try this

adriresu:
Last commit done to fit with #60

adriresu:
Reference code is not protected on xlsx exports
image

It is necessary to be protected? @ChristianSantamaria

mrp_unbuild_cust_alu_analytics/models/mrp_unbuild.py
adriresu: I tried with api.onchange, but looks like it doesn't work with one2many tree views

mrp_unbuild_cust_alu_analytics/views/mrp_bom_views.xml
adriresu: Hide

mrp_unbuild_cust_alu_analytics/views/mrp_unbuild_views.xml
adriresu: Hide

mrp_unbuild_cust_alu_analytics/models/mrp_unbuild.py
adriresu: Yes better as One2Many. Change done

mrp_unbuild_cust_alu_analytics/models/mrp_unbuild.py
adriresu: Done, It wasn't necesary

mrp_unbuild_cust_alu_analytics/models/mrp_unbuild.py
adriresu: Changed, It works locally

mrp_unbuild_cust_alu_analytics/models/mrp_unbuild.py
adriresu: Done

mrp_unbuild_cust_alu_analytics/models/mrp_unbuild_bom_quants.py
adriresu: Yes, removed at last commit

mrp_unbuild_cust_alu_analytics/models/mrp_unbuild_bom_quants.py
adriresu: Same, removed at last commit

mrp_unbuild_cust_alu_analytics/models/mrp_unbuild_bom_total.py
adriresu: Changed expected_qty to the 2 fields mentioned at the function

mrp_unbuild_cust_alu_analytics/security/ir.model.access.csv
adriresu: They can´t, limited process type and incidence type to mrp groups

mrp_unbuild_cust_alu_analytics/views/mrp_bom_views.xml
adriresu: Done

mrp_unbuild_cust_alu_analytics/views/mrp_unbuild_bom_total_views.xml
adriresu: Done

mrp_unbuild_cust_alu_analytics/views/mrp_unbuild_process_type_views.xml
adriresu: Done

mrp_unbuild_cust_alu_analytics/models/mrp_unbuild.py
```diff
@@ -0,0 +1,151 @@
...
+    process_type_id = fields.Many2one('mrp.unbuild.process.type', string='Process type')
+    incidence_ids = fields.One2many(
+        comodel_name="mrp.unbuild.incidence",
+        inverse_name="unbuild_id",
+        string="Incidences",
+    )
+    analytic_ids = fields.One2many(
+        comodel_name="qc.inspection",
+        string="Analytics",
+        store=False
```
dalonsod: `inverse_name` is missing (maybe is still linking data, but I don't really how it works). `store=False` does not make sense IMO. And you should add `copy=False`, in order to prevent copying analytics when duplicating an unbulld

mrp_unbuild_cust_alu_analytics/models/mrp_unbuild.py
dalonsod: If you see `super()` implementation:

https://github.com/odoo/odoo/blob/4cc086d330b73514fbc64a7fcf22d8a7a9f1b691/addons/mrp/models/mrp_unbuild.py#L106

it's already non-multi, so your code is coherent and there's no need to fix anything :+1:

mrp_unbuild_cust_alu_analytics/models/mrp_unbuild.py
dalonsod: Not agree. This code will be fired within a cron execution, not a user action. Who will catch those error and notice anyone. The best way should be wrapping it with try-except statement and logging errors.

@ChristianSantamaria what do you think?

mrp_unbuild_cust_alu_analytics/views/mrp_bom_views.xml
dalonsod: OK, then let's use `hide` instead of `0`.

mrp_unbuild_cust_alu_analytics/views/mrp_unbuild_views.xml
dalonsod: OK, then let's use `hide` instead of `0`.

mrp_unbuild_cust_alu_analytics/models/mrp_unbuild.py
adriresu: From what I have read, if any inverse_name is provided; it will link anyway but  the Many2one inverse field will not be created.
But yes, the correct is as u said, mostly to be bidirectional.

adriresu: Fixed

mrp_unbuild_cust_alu_analytics/models/mrp_unbuild.py
adriresu: I made a general try-except clause with nested try-except fired like this:
image

mrp_unbuild_cust_alu_analytics/views/mrp_bom_views.xml
adriresu: Done

mrp_unbuild_cust_alu_analytics/views/mrp_unbuild_views.xml
adriresu: Done

dalonsod (approved):
At last!!

PR #54, opened 2024-02-15
[14.0][IMP+FIX] fcd_weight_scale_mrp: kg limit and fix slowness

ChristianSantamaria:


ChristianSantamaria:
@dalonsod kg limit tested in test enviroment, the other commit is more difficult to test...

fcd_weight_scale_mrp/models/res_company.py
```diff
@@ -8,3 +8,4 @@ class ResCompany(models.Model):
 
     weight_scale_username_id = fields.Many2one('res.users',string='Username')
     weight_scale_password = fields.Char(string='Password')
+    weight_scale_limit = fields.Float(string='Limit Kg', default=200)
```
dalonsod: Remember this field belongs to company, so label should be more specific ("Limit kg" is very out of context). That simplified label has sense e.g. in res.config.settings form, because it will probable be placed in a section that talks a about scales configurations

fcd_weight_scale_mrp/models/res_conf_settings.py
```diff
@@ -13,3 +13,4 @@ class ResConfigSettings(models.TransientModel):
         readonly=False,
     )
     weight_scale_password = fields.Char(related='company_id.weight_scale_password', string="Password", readonly=False)
+    weight_scale_limit = fields.Float(related='company_id.weight_scale_limit', string="Limit", readonly=False)
```
dalonsod: If you remove sting attribute, it will be obtained from res.company field, not tested

dalonsod (changes requested):
See minor quality code comments

PR #55, opened 2024-02-21
[15.0][ADD] mrp_lot_manual

dalonsod:
Ensures that production lot is filled when production is intended to be marked as Done.

When a production is Marked as Done from Confirmed state, every consumption and production quantities are automatically filled, and lot/serial is automatically generated. In other cases when lot is not set yet an error is raised. This addon prevents that lost case, so lot is always manually filled.


dalonsod:
Validated by customer

PR #56, opened 2024-03-06
[13.0][IMP] mrp_unbuild_qc: shortcut buttons from unbuilds to moves inspections

dalonsod:
@adriresu @ChristianSantamaria  could you review?


ChristianSantamaria:
Installed and tested in alu test enviroment LGTM

PR #57, opened 2024-03-22
[13.0][ADD] mrp_unbuild_cust_alu_analytics_maintenance

adriresu:
Edit: depends on #53


mrp_unbuild_cust_alu_analytics_maintenance/__manifest__.py
```diff
@@ -0,0 +1,21 @@
...
+{
+    "name": "Unbuild Maintenance",
+    "summary": """
+        Bind unbuilds with maintenances
+    """,
+    "version": "13.0.1.0.0",
+    "category": "Manufacturing",
+    "website": "https://github.com/solvosci/slv-manufacture",
+    "author": "Solvos",
+    "license": "LGPL-3",
```
dalonsod: Wrong license, because this addon actually depends on OCA's `quality_control_oca`. It should be AGPL-3 (I know `mrp_unbuild_qc` is wrong licensed, I should fix it)

mrp_unbuild_cust_alu_analytics_maintenance/models/maintenance_equipment.py
```diff
@@ -0,0 +1,20 @@
...
+# License LGPL-3.0 (https://www.gnu.org/licenses/lgpl-3.0.html)
+
+from odoo import _, api, fields, models
+
+class MaintenanceEquipment(models.Model):
+    _inherit = "maintenance.equipment"
+
+    works_with_unbuild = fields.Boolean(string='Works with Unbuild', default=False)
+
+    unbuilds = fields.Integer(
```
dalonsod: Better naming:
```suggestion
    unbuild_count = fields.Integer(
```

mrp_unbuild_cust_alu_analytics_maintenance/models/maintenance_equipment.py
```diff
@@ -0,0 +1,20 @@
...
+        compute="_compute_count_unbuilds", string="Unbuild Incidences"
+    )
+
+    def _compute_count_unbuilds(self):
+        self.unbuilds = 0
+        for unbuild in self.env["mrp.unbuild"].search([]):
```
dalonsod: Please refactor this, we should never use a search without domain. There are some ways to improve this, e.g.:
- Use `read_group` from incidence model.
- Directly use `search` from incidence model.

Remember that a compute method is always multi, so this code won't work if it's called for more than a record.

mrp_unbuild_cust_alu_analytics_maintenance/views/maintenance_equipment_views.xml
```diff
@@ -0,0 +1,28 @@
+<odoo>
+    <data>
+        <record id="maintenance_equipment_total_unbuilds_form" model="ir.ui.view">
+            <field name="name">maintenance.equipment (mrp_total_unbuild)</field>
```
dalonsod: Better name:
```suggestion
            <field name="name">maintenance.equipment.form (mrp_total_unbuild)</field>
```

mrp_unbuild_cust_alu_analytics_maintenance/views/maintenance_equipment_views.xml
```diff
@@ -0,0 +1,28 @@
...
+            <field name="model">maintenance.equipment</field>
+            <field name="groups_id" eval="[(4, ref('mrp.group_mrp_user'))]"/>
+            <field name="inherit_id" ref="maintenance.hr_equipment_view_form"/>
+            <field name="arch" type="xml">
+                <xpath expr="//field[@name='owner_user_id']" position="after">
+                    <field name="works_with_unbuild"/>
```
dalonsod: Try `boolean_toggle` widget

mrp_unbuild_cust_alu_analytics_maintenance/views/maintenance_equipment_views.xml
```diff
@@ -0,0 +1,28 @@
...
+                    <field
+                        name="unbuilds"
+                        widget="statinfo"
+                        string="Unbuilds"
+                    />
+                </xpath>
+            </field>
+        </record>
+    </data>
+</odoo>
```
dalonsod: Missing empty EOF line

mrp_unbuild_cust_alu_analytics_maintenance/views/mrp_unbuild_views.xml
```diff
@@ -0,0 +1,28 @@
...
+        </record>
+
+        <record model="ir.actions.act_window" id="action_total_unbuilds">
+            <field name="name">Total Unbuilds</field>
+            <field name="res_model">mrp.unbuild</field>
+            <field name="view_mode">tree,form</field>
+            <field name="domain">[
+                ("id", "=", active_id),
+            ]</field>
+            <field name="context">{"hide_unbuild": True}</field>
```
dalonsod: This context value seems to be unused

mrp_unbuild_cust_alu_analytics_maintenance/views/mrp_unbuild_views.xml
```diff
@@ -0,0 +1,28 @@
+<odoo>
+    <data>
+        <record id="mrp_unbuild_form_incidence_ids_maintenance_equipment" model="ir.ui.view">
+            <field name="name">mrp.unbuild.form (maintenance_equipment) (mrp_unbuild_maintenance)</field>
+            <field name="model">mrp.unbuild</field>
+            <field name="inherit_id" ref="mrp_unbuild_cust_alu_analytics.mrp_unbuild_form_incidence_ids_analytics"/>
+            <field name="groups_id" eval="[(4,ref('maintenance.group_equipment_manager'))]"/>
```
dalonsod: This is too restricted. According to 

https://github.com/odoo/odoo/blob/13.0/addons/maintenance/security/ir.model.access.csv

every user can access a maintenance equipment, so we can remove such access without any problem.

dalonsod (changes requested):
See comments, and also:
- Add two complementary filters in search view for maintenance equipments: "enabled for unbuild incidences" and "disabled for unbuild incidences", only available for MRP users and equipment managers.

mrp_unbuild_cust_alu_analytics_maintenance/__manifest__.py
adriresu: Solved

mrp_unbuild_cust_alu_analytics_maintenance/models/maintenance_equipment.py
adriresu: Solved

mrp_unbuild_cust_alu_analytics_maintenance/models/maintenance_equipment.py
adriresu: Solved

mrp_unbuild_cust_alu_analytics_maintenance/views/maintenance_equipment_views.xml
adriresu: Solved

mrp_unbuild_cust_alu_analytics_maintenance/views/maintenance_equipment_views.xml
adriresu: I think this widget is not necesary, in odoo 13, boolean fields has boolean_toggle by default

mrp_unbuild_cust_alu_analytics_maintenance/views/maintenance_equipment_views.xml
adriresu: Solved

mrp_unbuild_cust_alu_analytics_maintenance/views/mrp_unbuild_views.xml
adriresu: Removed

mrp_unbuild_cust_alu_analytics_maintenance/views/mrp_unbuild_views.xml
adriresu: Removed

mrp_unbuild_cust_alu_analytics_maintenance/__manifest__.py
```diff
@@ -0,0 +1,21 @@
+# © 2024 Solvos Consultoría Informática (<http://www.solvos.es>)
+# License LGPL-3 - See http://www.gnu.org/licenses/lgpl-3.0.html
```
dalonsod: Change license comments according to new license (in this file and others)

mrp_unbuild_cust_alu_analytics_maintenance/models/maintenance_equipment.py
```diff
@@ -0,0 +1,19 @@
...
+
+    unbuild_count = fields.Integer(
+        compute="_compute_count_unbuilds", string="Unbuild Incidences"
+    )
+
+    def _compute_count_unbuilds(self):
+        for record in self:
+            incidence_ids = self.env["mrp.unbuild.incidence"].search([("maintenance_equipment_id", "=", record.id)])
+            unbuild_ids = incidence_ids.mapped("unbuild_id")
+            record.unbuild_count = len(unbuild_ids)
```
dalonsod, line 20: This code is right, but still inefficient. Since this field is only computed in a form, is fast (for a single record). But, if we should add it to e.g. a tree view it should become very slow.
Please simply add a TODO comment for a later improvement of this (we could replace it by a `read_group` strategy)

mrp_unbuild_cust_alu_analytics_maintenance/views/maintenance_equipment_views.xml
dalonsod: Is it currently rendering like this?

imagen

mrp_unbuild_cust_alu_analytics_maintenance/views/mrp_unbuild_views.xml
```diff
@@ -0,0 +1,26 @@
...
+        <record model="ir.actions.act_window" id="action_total_unbuilds">
+            <field name="name">Total Unbuilds</field>
+            <field name="res_model">mrp.unbuild</field>
+            <field name="view_mode">tree,form</field>
+            <field name="domain">[
+                ("incidence_ids.maintenance_equipment_id", "=", active_id),
+            ]</field>
+        </record>
+    </data>
+</odoo>
```
dalonsod: Missing EOF empty line

dalonsod (commented):
See new comments. I've finished code review, I'll make functional review ASAP

mrp_unbuild_cust_alu_analytics_maintenance/__manifest__.py
adriresu: Done

mrp_unbuild_cust_alu_analytics_maintenance/models/maintenance_equipment.py
adriresu: Should I use read_group allways for this cases, since with search function, it never will be efficient enought with multiple records?

mrp_unbuild_cust_alu_analytics_maintenance/views/mrp_unbuild_views.xml
adriresu: Done

mrp_unbuild_cust_alu_analytics_maintenance/models/maintenance_equipment.py
dalonsod: For non-stored count compute fields, the rule should be as follows:
- if count is only accessed within a form view, "len()" strategy should be enough. Or `search_count()` as well. In both cases, compute method is called with a single record, so no matter of the strategy, it should be fast (but be careful, a form access could be only in early development status and later added other access that make better the below strategy).
- if count is accessed from other views (tree, kanban, and others), `read_group()` should be the best option. There are a lot of examples in OCA/solvosci.

mrp_unbuild_cust_alu_analytics_maintenance/views/maintenance_equipment_views.xml
adriresu: After new tests, widget is neccesary to render like this

adriresu:
From my side, ready to merge.

mrp_unbuild_cust_alu_analytics_maintenance/views/mrp_unbuild_views.xml
```diff
@@ -0,0 +1,26 @@
+<odoo>
+    <data>
+        <record id="mrp_unbuild_form_incidence_ids_maintenance_equipment" model="ir.ui.view">
+            <field name="name">mrp.unbuild.form (maintenance_equipment) (mrp_unbuild_maintenance)</field>
+            <field name="model">mrp.unbuild</field>
+            <field name="inherit_id" ref="mrp_unbuild_cust_alu_analytics.mrp_unbuild_form_incidence_ids_analytics"/>
+            <field name="arch" type="xml">
+                <xpath expr="//page/field[@name='incidence_ids']/tree/field[@name='priority']" position="after">
```
ChristianSantamaria: field "priority" not exist now
```suggestion
                <xpath expr="//page/field[@name='incidence_ids']/tree/field[@name='machine']" position="after">
```

ChristianSantamaria (changes requested):

dalonsod (approved):
:+1:

PR #59, opened 2024-04-10
[13.0][IMP] mrp_unbuild_qc+mrp_unbuild_advanced_qc: backend inspections only linked to unbuilds support

dalonsod:
With this improvement it's possible to create quality inspections linked to unbuilds without selecting a specific unbuild move.

This is only possible from backend at this time.


dalonsod:
@ChristianSantamaria @adriresu with this improvement it's no longer required creating a QC inspection linked to an unbuild AND a stock move linked to that unbuild. A consequence of this is that such unbuild could be in a draft state without problem.

Then, in #53, in inspections import process:
- we should remove unbuild finished check, and
- when creating the inspection stock move determination becomes unnecessary, because `object_id` value is no longer required (unbuild stock moves couldn't exist at this point indeed).
- Product for inspection will be automatically filled with unbuild main product (thar could not be the right one, but we're aware of this, at this point).

We'll later add final inspection product to the inspection (and later linked with move when unbuild is done), but not for now.

adriresu:
Updated #53  on next to last commit.
Tested locally and it works 👍

dalonsod:
Good, ready to be merged!

PR #60, opened 2024-04-10
[13.0][IMP] mrp_unbuild_bom_cust_qty: total unbuild product qty automation // units coverage

dalonsod:
Edit:
With this improvement it's possible to automatically update unbuild
product quantity with the sum of every BoM quants, depending on a
general parameter that could be (de)activated.

And also, due to quantity calculation, units become important, starting
from this commit are shown in bom detailed and total lists.
Those quantities that belong to a category unit that is different than
product unbuild one are ignored during main quantity calculation and
shown in italics.

cc @ChristianSantamaria @adriresu


dalonsod:
@ChristianSantamaria @adriresu could you test?

With this improvement,
- Total unbuilt quantity is automatically calculated, and taking in account proper component units => **such calculation must be removed in #53** 
- Units are shown (when manage multiple units is enabled).
- BoM lines with a category unit that are different than product unbuild one are shown in italics:

imagen

adriresu:
Tested locally with #53 changes and worked as expected 👍

dalonsod:
Great! Merging it... if something is missing and/or a later bug is detected, I'll made another PR

PR #61, opened 2024-04-12
[13.0][FIX] mrp_unbuild_bom_cust_qty: ignore service products when unbuild is done

dalonsod:
This fix affect to a general bug in Odoo (that already enable adding services to a BoM, and thus, create stock moves with them once unbuild is finished), but also another generated by this addon, for custom BoM lines added by users for a certain unbuild.

cc @ChristianSantamaria


dalonsod:
Rebased in order to add #60

dalonsod:
Not tested yet, please rebase #53 after this PR is merged and check It in AluTest environment

PR #62, opened 2024-04-16
[13.0][IMP] mrp_unbuild_force_exact_qty: change exact quantity policy to always same UoM-categ products

dalonsod:
With this improvement, quantity check is always made, even if there are produced lines with different UoM category. Those moves are ignored anyway. In previous version, quantity check was disabled in that case.
    
Produced vs consumed moves filtering is improved, now are separated depending on destination location usage, instead of product definition.


dalonsod:
@ChristianSantamaria this improvement is now required if we can properly cover this two new possible situations:
* New produced lines with service products (in general, products whose UoM category doesn't belong to consumed one) => in previous version quantity check was simply ignored.
* The same product that was consumed was also produced => quantity check calculation had a bug in that case.

Could you check that in both cases this addon works as expected?

ChristianSantamaria:
LGTM :+1:

PR #63, opened 2024-04-16
[13.0][IMP] mrp_unbuild_bom_cust_qty: enabled adding unbuild product as component

dalonsod:
This is achieved manually adding the product, as custom BoM line, during unbuild creation (and later when state is still draft) and product is already present as a BoM line for at least one BoM. If not, an instrumental BoM is created in that moment, that can be used in later unbuilds with the same product and behavior.


dalonsod:
@ChristianSantamaria ready to be tested in AluTest environment

dalonsod:
Successfully tested in AluTest environment, ready to be merged

PR #66, opened 2024-04-22
[13.0][ADD] mrp_unbuild_view_improves

adriresu:


mrp_unbuild_view_improves/__manifest__.py
```diff
@@ -0,0 +1,18 @@
+# © 2024 Solvos Consultoría Informática (<http://www.solvos.es>)
+# License AGPL-3 - See http://www.gnu.org/licenses/lgpl-3.0.html
```
dalonsod: Wrong license comment

mrp_unbuild_view_improves/__manifest__.py
```diff
@@ -0,0 +1,18 @@
...
+    "name": "MRP Unbuild View Improves",
+    "summary": """
+        Adds some improvements for Unbuild views
+    """,
+    "author": "Solvos",
+    "license": "LGPL-3",
+    "version": "13.0.1.0.0",
+    "category": "Manufacturing",
+    "website": "https://github.com/solvosci/slv-manufacture",
+    "depends": ["mrp", "mrp_unbuild_advanced"],
```
dalonsod, line 13: `mrp_unbuild_advanced` dependency should be avoided. It's the opposite situation: we should consider moving some current `mrp_unbuild_advanced` functionality to this new addon, and make `mrp_unbuild_advanced` dependent of this new one, if it was finally required (and only in that case). Please comment if you don't understand philosophy behind it.

mrp_unbuild_view_improves/views/mrp_unbuild_views.xml
```diff
@@ -0,0 +1,21 @@
...
+                    <field name="name"/>
+                    <!-- <field name="process_type_id"/> -->
+                    <field name="product_id"/>
+                </calendar>
+            </field>
+        </record>
+
+        <!-- Action -->
+        <record id="mrp.mrp_unbuild" model="ir.actions.act_window">
+            <field name="view_mode">tree,kanban,form,calendar</field>
```
dalonsod, line 20: Since this is actually an action record update, please add a comment with old content, so you can visually see which exactly is the change made.

dalonsod (changes requested):
See comments

mrp_unbuild_view_improves/__manifest__.py
adriresu: Fixed, as AGPL

mrp_unbuild_view_improves/__manifest__.py
adriresu: Calendar view has unbuild_date field as date_start which is a field defined on mrp_unbuild_advanced.
How can we bypass this dependency?

mrp_unbuild_view_improves/views/mrp_unbuild_views.xml
adriresu: Should I add the entire action or only just the attribute affected?

dalonsod: Only affected attribute

mrp_unbuild_view_improves/__manifest__.py
dalonsod: Right one is LGPL-3. We'll only use AGPL-3 if an OCA addon with such license is involved

mrp_unbuild_view_improves/views/mrp_unbuild_views.xml
adriresu: Done

mrp_unbuild_view_improves/__manifest__.py
dalonsod: Ooops you're right. I'll think a little bit about it

mrp_unbuild_view_improves/__manifest__.py
adriresu: Ok, noted and updated

adriresu:
I added bom_code to calendar view, by adding a related field bom_code.
I think there is no other option to add a field of other model to the calendar view.

dalonsod:
That's right, in any view ya can only access fields of current model, so defining it is needed

mrp_unbuild_view_improves/models/mrp_unbuild.py
```diff
@@ -0,0 +1,15 @@
...
+from odoo import _, api, models, fields
+from odoo.exceptions import ValidationError
+
+
+class MrpUnbuild(models.Model):
+    _inherit = "mrp.unbuild"
+
+    bom_code = fields.Char(
+        string="Bom Code",
+        readonly=True,
```
dalonsod: In v13 a related field is read only by default, so you can remove attribute

mrp_unbuild_view_improves/models/mrp_unbuild.py
```diff
@@ -0,0 +1,15 @@
+# © 2024 Solvos Consultoría Informática (<http://www.solvos.es>)
+# License LGPL-3.0 (https://www.gnu.org/licenses/lgpl-3.0.html)
+
+from odoo import _, api, models, fields
+from odoo.exceptions import ValidationError
```
dalonsod: Please check, some of the imports are unused

adriresu: Done

mrp_unbuild_view_improves/__manifest__.py
dalonsod: Due to this problem, we've finally chosen moving it to `mrp_unbuild_advanced` as a PR, then created #69

PR #64, opened 2024-04-19
[13.0][IMP] mrp_unbuild_cust_alu_analytics: some improvements (total qty in report, times format, related product)

adriresu:
Edit: 
* Reuse get product_qty method from mrp_unbuild_bom_cust_qty, and add it in manifest.
* Set break_time and stop_time from report, to time format.
* Add rel_unbuild_product_id value after read XML.

PR #67, opened 2024-04-22
[13.0][IMP] mrp_unbuild_qc: related unbuild product support

dalonsod:
With this improvement, an additional product for a QC inspection is added and can be manually stablished and/or automatically filled depending on linked stock move.
This could be useful for those inspections programmatically created and not yet linked to an unbuild stock move (e.g. when unbuild is still in a draft state).

@adriresu now it's possible linking to custom unbuild product component in `mrp_unbuild_cust_alu_analytics` during inspection import process, simply adding `rel_unbuild_product_id` value calling `create()` method for `qc.inspection`, could you try? cc @ChristianSantamaria

### DIFF
--- a/mrp_unbuild_qc/__manifest__.py
+++ b/mrp_unbuild_qc/__manifest__.py
@@ -7,7 +7,7 @@
     """,
     "author": "Solvos",
     "license": "LGPL-3",
-    "version": "13.0.2.0.0",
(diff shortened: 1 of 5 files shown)


dalonsod:
Not fully tested yet, but merge is required before rebasing in #64 . Let's go...

PR #68, opened 2024-04-23
[13.0][IMP] mrp_unbuild_advanced: Allow future unbuild dates, when unbuild …

adriresu:
…isn't confirmed


mrp_unbuild_advanced/models/mrp_unbuild.py
```diff
@@ -68,6 +63,8 @@ def action_validate(self):
                     "Shift dates must be filled before"
                     " unbuild action for %s is performed!"
                 ) % unbuild.name)
+            if self.unbuild_date > fields.Datetime.now():
```
dalonsod: I don't understand why this check is placed here. According to requirements (check if date is still in the future when unbuild is marked as done), check should be made when "Unbuild button" is clicked:

https://github.com/odoo/odoo/blob/0d5b386c32db609d9a1d07ffe515d77fc83c3786/addons/mrp/models/mrp_unbuild.py#L137

Could you check?

dalonsod (changes requested):

mrp_unbuild_advanced/models/mrp_unbuild.py
adriresu: Check moved, but just to note, in case of insufficient qty, error will be triggered after confirm insufficient qty wizard, this may be, not much, but a bit tedious for the end user.

adriresu:
Translation is needed, after all changes are complete

mrp_unbuild_advanced/models/mrp_unbuild.py
dalonsod: OK, good point. Then try placing it in `action_validate()` method instead (very "Unbuild" button next step).

adriresu: Done

dalonsod: You're using `self`, but `action_validate()` is multi, please fix it

adriresu: Yeah, mb. Fixed

dalonsod (approved):
Not tested (we'll directly do it in AluTest environment), but it seems to be :+1: 

Please do the following operations:
- Add pending translation.
- Squash commits, but
- place README.rst and index.html files in a separate commit, like https://github.com/solvosci/slv-manufacture/pull/63/commits. Preserve CONTRIBUTORS.rst in squashed commit.

ChristianSantamaria:
LGTM :+1:

PR #70, opened 2024-04-25
[13.0][IMP] stock_valuation_mrp: Add no_manufacture_unbuild_valuation, cost…

adriresu:
…_management_residue, field and views

### DIFF
--- a/stock_valuation_mrp/models/mrp_bom_line.py
+++ b/stock_valuation_mrp/models/mrp_bom_line.py
@@ -6,7 +6,7 @@ from odoo import models, fields
 class MrpBomLine(models.Model):
     _inherit = "mrp.bom.line"
 
-    no_manufacture_unbuild_valuation = fields.Boolean(
-        string='No valuation for manufacturing/unbuild',
+    disabled_mrp_unbuild_valuation = fields.Boolean(
+        string='Disabled for manufacturing/unbuild',
         default=False
     )
--- a/stock_valuation_mrp/models/mrp_bom_line.py
+++ b/stock_valuation_mrp/models/mrp_bom_line.py
@@ -1,0 +1,12 @@
+# © 2024 Solvos Consultoría Informática (<http://www.solvos.es>)
+# License LGPL-3.0 (https://www.gnu.org/licenses/lgpl-3.0.html)
+
+from odoo import models, fields
+
+class MrpBomLine(models.Model):
+    _inherit = "mrp.bom.line"
+
+    no_manufacture_unbuild_valuation = fields.Boolean(
+        string='No valuation for manufacturing/unbuild',
+        default=False
+    )
--- a/stock_valuation_mrp/models/product_product.py
+++ b/stock_valuation_mrp/models/product_product.py
@@ -1,0 +1,22 @@
+# © 2024 Solvos Consultoría Informática (<http://www.solvos.es>)
+# License LGPL-3.0 (https://www.gnu.org/licenses/lgpl-3.0.html)
+
+from odoo import models, fields, api
+
+class ProductProduct(models.Model):
+    _inherit = "product.product"
+
+    cost_management_residue = fields.Boolean(
+        string='Has cost management residue',
+        default=False,
+        tracking=True,
+        help="""
+        When enabled, product residue will be covered with cost management
+        """,
+    )
+
+    @api.depends('product_category_id.warehouse_valuation')
+    def _compute_warehouse_valuation(self):
+        for record in self:
+            if not record.categ_id.warehouse_valuation:
+                record.cost_management_residue = False
--- a/stock_valuation_mrp/models/product_product.py
+++ b/stock_valuation_mrp/models/product_product.py
@@ -6,17 +6,53 @@ from odoo import models, fields, api
 class ProductProduct(models.Model):
     _inherit = "product.product"
 
-    cost_management_residue = fields.Boolean(
+    has_waste_cost_mgmt = fields.Boolean(
         string='Has cost management residue',
         default=False,
         tracking=True,
+        readonly=False,
+        compute='_compute_has_waste_cost_mgmt',
+        store=True,
         help="""
         When enabled, product residue will be covered with cost management
         """,
     )
 
-    @api.depends('product_category_id.warehouse_valuation')
-    def _compute_warehouse_valuation(self):
-        for record in self:
-            if not record.categ_id.warehouse_valuation:
-                record.cost_management_residue = False
+    residue_pricelist_mgmt = fields.Many2one(
+        comodel_name='product.pricelist',
+        string='Pricelist for management residue',
+        default=False,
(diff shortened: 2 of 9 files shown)


stock_valuation_mrp/models/mrp_bom_line.py
```diff
@@ -0,0 +1,12 @@
...
+
+from odoo import models, fields
+
+class MrpBomLine(models.Model):
+    _inherit = "mrp.bom.line"
+
+    no_manufacture_unbuild_valuation = fields.Boolean(
+        string='No valuation for manufacturing/unbuild',
+        default=False
+    )
```
dalonsod: Better naming IMO:
```suggestion
    disabled_mrp_unbuild_valuation = fields.Boolean(
        string='Disabled for manufacturing/unbuild valuation',
        default=False
    )
```

stock_valuation_mrp/models/product_product.py
```diff
@@ -0,0 +1,52 @@
...
+        tracking=True,
+        readonly=False,
+        compute='_compute_has_waste_cost_mgmt',
+        store=True,
+        help="""
+        When enabled, product residue will be covered with cost management
+        """,
+    )
+
+    residue_pricelist_mgmt = fields.Many2one(
```
dalonsod: Make this field name coherent with related check one

stock_valuation_mrp/models/product_product.py
```diff
@@ -0,0 +1,52 @@
...
+        comodel_name='product.pricelist',
+        string='Pricelist for management residue',
+        default=False,
```
dalonsod: This is unnecessary (it's relevant for `Boolean` fields, but for `Many2one` field is obviously the default value...)

stock_valuation_mrp/models/product_product.py
```diff
@@ -0,0 +1,52 @@
...
+        store=True,
+        help="""
+        Pricelist used for cost management residue
+        """,
+    )
+
+    @api.depends("categ_id.warehouse_valuation")
+    def _compute_has_waste_cost_mgmt(self):
+        for product in self:
+            product.has_waste_cost_mgmt = False # Allways False if warehouse_valuation change (True or False)
```
dalonsod: Could you try a more multi way for this value setting? And there's no need to set value to `False` if current value is already `False`, so a filtered for non `False` values should be added, shouldn't it?

stock_valuation_mrp/models/product_product.py
```diff
@@ -0,0 +1,52 @@
...
+
+
+    @api.depends("has_waste_cost_mgmt")
+    def _compute_residue_pricelist_mgmt(self):
+        for product in self:
+            product.residue_pricelist_mgmt = False # Same as above
```
dalonsod: The same as above

stock_valuation_mrp/models/product_template.py
```diff
@@ -0,0 +1,83 @@
...
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        templates = super().create(vals_list)
+        for template, vals in zip(templates, vals_list):
+            related_vals = {}
+            if vals.get("has_waste_cost_mgmt"):
+                related_vals["has_waste_cost_mgmt"] = vals["has_waste_cost_mgmt"]
+            if vals.get("residue_pricelist_mgmt"):
+                related_vals["residue_pricelist_mgmt"] = False
```
dalonsod: Why this value is forced to `False`?

stock_valuation_mrp/views/product_template_views.xml
```diff
@@ -0,0 +1,32 @@
...
+                />
+                <field
+                    name="residue_pricelist_mgmt"
+                    attrs="{
+                        'invisible': 
+                        [
+                            '|',
+                                ('has_waste_cost_mgmt','=',False),
+                                ('warehouse_valuation','=',False)
+                        ],
```
dalonsod, line 24: Why this condition is different than product variant form one? `has_waste_cost_mgmt` checking should be enough, if I'm right...

dalonsod (changes requested):
See comments

stock_valuation_mrp/models/product_product.py
adriresu: Maybe residue_mgmt_pricelist_id is a good name?

dalonsod: I'd refer to use "waste" instead of "residue"

stock_valuation_mrp/views/product_template_views.xml
adriresu: Changing the warehouse_valuation from a product category might cause has_waste_cost_mgmt to become True even if warehouse_valuation itself is False.
Resulting in a view with has_waste_cost_mgmt as invisible, and pricelist as visible

stock_valuation_mrp/models/mrp_bom_line.py
adriresu: Changed

stock_valuation_mrp/models/product_product.py
adriresu: Changed to waste_mgmt_pricelist_id

stock_valuation_mrp/models/product_product.py
adriresu: Removed

stock_valuation_mrp/models/product_product.py
adriresu: Done

stock_valuation_mrp/models/product_product.py
adriresu: Done

stock_valuation_mrp/models/product_template.py
adriresu: It's a mistake, updated to vals[residue_pricelist_mgmt]

stock_valuation_mrp/models/product_product.py
```diff
@@ -0,0 +1,55 @@
...
+        Pricelist used for cost management residue
+        """,
+    )
+
+    @api.depends("categ_id.warehouse_valuation")
+    def _compute_has_waste_cost_mgmt(self):
+        valuation_products = self.filtered(lambda x: x.warehouse_valuation == False and x.has_waste_cost_mgmt == True)
+        valuation_products.write(
+            {"has_waste_cost_mgmt": False}
+        )
```
dalonsod, line 38: OK, don't change anything, only for a similar situation later, try simpler code:
```suggestion
        self.filtered(lambda x: not x.warehouse_valuation and x.has_waste_cost_mgmt).write({
            "has_waste_cost_mgmt": False
        })
```

dalonsod:
:+1: I'll test it ASAP

dalonsod:
Code moved to new addon `mrp_unbuild_cust_alu_analytics_valuation` for better performance and simplicity. Close this.

PR #71, opened 2024-04-26
[13.0][IMP+FIX] mrp_unbuild_cust_alu_analytics[_maintenance]: Tablet support to operators // Ensure access to maintenance equipments

adriresu:
…ss.type model


mrp_unbuild_cust_alu_analytics/views/mrp_unbuild_process_type_views.xml
```diff
@@ -11,6 +11,7 @@
                             <group>
                                 <field name="name"/>
                                 <field name="cmplanta_id"/>
+                                <field name="company_id" invisible="True"/>
```
dalonsod: If company becomes invisible, we could never select it when requiered (e.g. in a multi-company scenario). It's not our current situation (customer doesn't work with multi-company), but is not right. Try this instead (https://github.com/odoo/odoo/blob/4cc086d330b73514fbc64a7fcf22d8a7a9f1b691/addons/mrp/views/mrp_bom_views.xml#L70C29-L70C119):
```suggestion
                                <field name="company_id" groups="base.group_multi_company" options="{'no_create': True}"/>
```

User won't see company field if it's working in a single-company scenario, but in a hypothetical future scenario with two or more companies it will be shown.

mrp_unbuild_cust_alu_analytics/models/mrp_unbuild_process_type.py
```diff
@@ -7,6 +7,11 @@ class MrpUnbuildProcessType(models.Model):
     _name = "mrp.unbuild.process.type"
     _description = 'Unbuild Process Type'
 
+    company_id = fields.Many2one(
+        comodel_name="res.company",
+        default=lambda self: self.env.company,
+        store=True,
```
dalonsod: `store` attribute has no sense (this is a direct field, not a compute one), please remove it.

mrp_unbuild_cust_alu_analytics/models/mrp_unbuild_process_type.py
```diff
@@ -7,6 +7,11 @@ class MrpUnbuildProcessType(models.Model):
     _name = "mrp.unbuild.process.type"
     _description = 'Unbuild Process Type'
 
+    company_id = fields.Many2one(
+        comodel_name="res.company",
+        default=lambda self: self.env.company,
```
dalonsod: add index for the field, stuff required for a good company access rule performance
```suggestion
        default=lambda self: self.env.company,
        index=True,
```

dalonsod (changes requested):
See comments.
And I miss the required global company record rule for this model, like 

https://github.com/odoo/odoo/blob/4cc086d330b73514fbc64a7fcf22d8a7a9f1b691/addons/mrp/security/mrp_security.xml#L67-L72

dalonsod:
> And I miss the required global company record rule for this model, like
> 
> https://github.com/odoo/odoo/blob/4cc086d330b73514fbc64a7fcf22d8a7a9f1b691/addons/mrp/security/mrp_security.xml#L67-L72

Still missing this 👆

mrp_unbuild_cust_alu_analytics/views/mrp_unbuild_process_type_views.xml
adriresu: Done

mrp_unbuild_cust_alu_analytics/models/mrp_unbuild_process_type.py
adriresu: Removed

mrp_unbuild_cust_alu_analytics/security/mrp_unbuild_cust_alu_analytics_security.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <data noupdate="0">
```
dalonsod: Record rules should be always defined in the way admin user can change them:
```suggestion
    <data noupdate="1">
```

If you take a look again at the provided example you could realize that rules are under a `noupdate="1"` section, and group ones are not. See more at https://github.com/solvosci/slv-docker-odoo-prod/wiki/Odoo-SLV-contribution-brief-guidelines#5-security

dalonsod (commented):
Make required change, anyway I'll test it ASAP

adriresu:
Last commit, where primary view is inheriting, is not working yet.
Looks like It has some problems inheriting primary view as another primary view, or maybe is related with the primary view which is inherited cuz it priority is 9999
[](https://github.com/solvosci/slv-manufacture/blob/13.0/mrp_unbuild_bom_cust_qty/views/mrp_unbuild_bom_quants.xml#L31)

adriresu:
> Last commit, where primary view is inheriting, is not working yet. Looks like It has some problems inheriting primary view as another primary view, or maybe is related with the primary view which is inherited cuz it priority is 9999

[https://github.com/solvosci/slv-manufacture/blob/13.0/mrp_unbuild_bom_cust_qty/views/mrp_unbuild_bom_quants.xml#L31](url)

dalonsod:
> Last commit, where primary view is inheriting, is not working yet. Looks like It has some problems inheriting primary view as another primary view, or maybe is related with the primary view which is inherited cuz it priority is 9999

I don't understand which the problem is. I've looked for references to new `mrp_unbuild_bom_quants_view_form_extended_only_save` view and didn't find them. A new primary view (and with lower priority) is well defined if we want to keep current behavior (the original view will be still called, without any change). Then, such new view will only be called if an action explicity call it...

dalonsod:
Last changes:

- Added a last commit for maintenance equipment proper access. Users that are not maintenance administrators  and are not followers of an equipment couldn't access it. Fixed (every MRP User can access them for now, for those that have the proper mark checked).
- Renamed PR according full functionality provided.

@ChristianSantamaria @adriresu please notice that this PR adds a new model (`mrp.incidence.description`) that has no `company_id`, the same case that e.g. `mrp.incidence.type` and `mrp.unbuild.incidence`. Please remember that when we are creating a new model we should ask ourselves about it and add it when needed. By the way, we'll keep them without company.

PR #81, opened 2024-08-01
[13.0][IMP] mrp_unbuild_cust_alu_analytics: ignore unbuild no-named .xml analytics

dalonsod:
When unbuild is not found, there are two possibilities:
- No name found => it's a test analytic, so it should be marked as processed, like a right one.
- Incorrect unbuild name => it should be kept as unprocessed, so external users can check it and fix the problem.

cc @ChristianSantamaria


dalonsod:
Tested in AluTest environment :+1:

PR #82, opened 2024-08-05
[13.0][FIX] mrp_unbuild_cust_alu_analytics: wrong onchange product_id

ChristianSantamaria:
Error when selecting a producto with several BOMs

### DIFF
--- a/mrp_unbuild_cust_alu_analytics/models/mrp_unbuild.py
+++ b/mrp_unbuild_cust_alu_analytics/models/mrp_unbuild.py
@@ -27,7 +27,7 @@ class MrpUnbuild(models.Model):
     @api.onchange('product_id')
     def _onchange_product_id(self):
         if self.product_id and self.process_type_id:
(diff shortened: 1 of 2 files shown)


mrp_unbuild_cust_alu_analytics/models/mrp_unbuild.py
```diff
@@ -27,7 +27,7 @@ class MrpUnbuild(models.Model):
     @api.onchange('product_id')
     def _onchange_product_id(self):
         if self.product_id and self.process_type_id:
-            self.bom_id = self.product_id.product_tmpl_id.bom_ids.filtered(lambda r: r.process_type_id == self.process_type_id)
+            self.bom_id = self.product_id.product_tmpl_id.bom_ids.filtered(lambda r: r.process_type_id == self.process_type_id)[0]
```
dalonsod, line 30: Origin code didn't fail in one case: when there wasn't any BoM for the selected process type (as filtered returned `False` and, then, simply assign `False`). New code could fail in that case (because `False[0]` will raise an error), please protect it.

dalonsod (changes requested):

PR #84, opened 2024-09-17
[15.0][ADD] mrp_production_cancel_manager

IriaAlonso:
Only allow to cancel productions to responsible manufacturing users


mrp_production_cancel_manager/views/mrp_production_view.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+
+    <record id="mrp_production_form" model="ir.ui.view">
+        <field name="name">mrp.production.form</field>
+        <field name="model">mrp.production</field>
+        <field name="inherit_id" ref="mrp.mrp_production_form_view"/>
+        <field name="arch" type="xml">
+            <xpath expr="//button[@name='action_cancel']" position="attributes">
```
dalonsod, line 9: This code seems to be OK, but it might be incomplete, because there are already two Cancel buttons, and that xpath expression matches both:

https://github.com/odoo/odoo/blob/64a4ca83cd9b2f208c555942346bb62e31825266/addons/mrp/views/mrp_production_views.xml#L113-L117

Could you check if this code applies to both buttons, or only the first or last one?

IriaAlonso:
@dalonsod do you like it? I tried other but they didn't work.

dalonsod:
Not sure, I'll take a look...

dalonsod:
Reviewed in CrtTest environment, this implementation is enough :+1: Let's merge...

PR #89, opened 2024-10-25
[13.0][FIX] mrp_unbuild_cust_alu_analytics_valuation: process types moved from hour to unit based

dalonsod:
With this fix, amortization and repair/maintenance unit costs are treated like consumables or maquilas (based on unit, e.g. "tons") rather than hours based.


dalonsod:
Approved by customer, merging...

PR #90, opened 2024-11-07
[13.0][IMP] mrp_unbuild_bom_cust_qty + mrp_unbuild_cust_alu_analytics_valuation: unbuild lots coverage + output valuation impact

dalonsod:
Two improvements:

- `mrp_unbuild_bom_cust_qty`: _generated products lots coverage_. With this improvement, now it's possible to add final product quantities with lots, when product is tracked. These modifications are inspired in OCA's `mrp_unbuild_tracked_raw_material`, which bypasses stantard process regarding to tracked products (that are forbidden by default), unfortunately forcing some copy & past code.

- `mrp_unbuild_cust_alu_analytics_valuation`: _output pickings valuation data_. When an output picking is done some valuation data is added when stock moves are linked to unbuilds via used lots. Otherwise, valuation info will use average price for the proper date, warehouse and product.


dalonsod:
:+1: in AluTest environment and approved by customer.

PR #92, opened 2024-11-14
[15.0][IMP] mrp_project_analytic: mark analytic_mrp_from_child with purchase.order.line

ChristianSantamaria:
Adds mark analytic_mrp_from_child if exist purchase.order.line with same product and account.analytic


mrp_project_analytic/models/mrp_production.py
```diff
@@ -40,6 +40,14 @@ def mark_analytic_mrp_from_child(self):
         )
         aml_ids = raw_material_moves.sudo().mapped('account_move_ids.line_ids')
         
+        for move in self.move_raw_ids.filtered(lambda x: x.state == "done"):
+            mr_qty = sum(self.search([('project_id', '=', self.project_id.id), ('state', '=', 'done')]).move_raw_ids.filtered(lambda x: x.product_id == move.product_id.id).mapped('quantity_done'))
```
dalonsod: `self` is could be `multi`, because `button_mark_done` is `multi`, so `self.project_id.id` could be raise an error. Such call should be replaced by "get the project from the production of current move" instead.

mrp_project_analytic/models/mrp_production.py
```diff
@@ -40,6 +40,14 @@ def mark_analytic_mrp_from_child(self):
...
+            mt_total = mr_qty + move.product_uom_qty
+            pol_qty = sum(self.env['purchase.order.line'].search([('account_analytic_id', '=', self.project_id.analytic_account_id.id), ('state', 'in', ['purchase', 'done']), ('product_id', '=', move.product_id.id)]).mapped('product_uom_qty'))
```
dalonsod: The same as above for `self.project_id`.

dalonsod: `mrp_project_analytic` doesn't depend on `purchase`. We should consider adding `purchase` or `purchase_mrp` as dependency, or hooking this module and creating another one with this piece of code that actually depends on any purchase addon.

dalonsod (changes requested):
See comments, but we'll wait for pending meeting with LR about this improvement

mrp_project_analytic_purchase/models/mrp_production.py
```diff
@@ -0,0 +1,22 @@
...
+    _inherit = "mrp.production"
+
+    def _prepare_mark_account_move_ids(self):
+        res = super(ManufactureOrder, self)._prepare_mark_account_move_ids()
+
+        for move in self.move_raw_ids.filtered(lambda x: x.state == "done"):
+            # TODO: Check slow query
+            mr_qty = sum(self.sudo().search([('project_id', '=', move.raw_material_production_id.project_id.id), ('state', '=', 'done')]).move_raw_ids.filtered(lambda x: x.product_id == move.product_id.id).mapped('quantity_done'))
+            mt_total = mr_qty + move.product_uom_qty
+            pol_qty = sum(self.env['purchase.order.line'].sudo().search([('account_analytic_id', '=', move.raw_material_production_id.project_id.analytic_account_id.id), ('state', 'in', ['purchase', 'done']), ('product_id', '=', move.product_id.id)]).mapped('product_uom_qty'))
```
dalonsod: Only performance suggested improvement (not tested) in order to prevent repeated calculations (`sudo` purchase lines object, and move project):
```suggestion
       pol_obj = self.env['purchase.order.line'].sudo()
        for move in self.move_raw_ids.filtered(lambda x: x.state == "done"):        
            project_id = move.raw_material_production_id.project_id
            # TODO: Check slow query
            mr_qty = sum(self.sudo().search([('project_id', '=', project_id.id), ('state', '=', 'done')]).move_raw_ids.filtered(lambda x: x.product_id == move.product_id.id).mapped('quantity_done'))
            mt_total = mr_qty + move.product_uom_qty
            pol_qty = sum(pol_obj.search([('account_analytic_id', '=', project_id.analytic_account_id.id), ('state', 'in', ['purchase', 'done']), ('product_id', '=', move.product_id.id)]).mapped('product_uom_qty'))
```

dalonsod (approved):
:+1:

PR #93, opened 2024-11-20
[15.0][IMP] mrp_production_buttons_manager: add confirmation when a production is cancelled

IriaAlonso:
- Add confirmation when a production is cancelled

@dalonsod


dalonsod:
OK, I'll test it ASAP.

Meanwhile, split commit into two:
- First commit with the current title, with whole functionality, but removing two files (`README.rst` addon root file update, and `index.html`).
- Second commit, with these two pending files, and this title: `[UPD] mrp_production_buttons_manager: README & index`.

IriaAlonso:
@dalonsod done

dalonsod:
Seems to be wrong, if you see commit details, you'll figure out that first commit has completely removed both files, and second one added them completely.

We should have this behavior instead:
- First commit shouldn't have any changes related to these two files.
- Second commit should only have files changes.

IriaAlonso:
@dalonsod

dalonsod (approved):
Tested in CrtTest environment :+1:

PR #94, opened 2024-12-13
[13.0][FIX] mrp_unbuild_cust_alu_analytics_valuation: sudo computation for unitary costs

dalonsod:
With this fix accessing affected fields by `_compute_cost_unitary_fields()` method is granted, preventing future record rules for unbuilds that could crash with older implementation (a write access error could be raised even loading the form).


dalonsod:
Direct merge

PR #96, opened 2025-01-14
[13.0][IMP] mrp_unbuild_cust_alu_analytics: adds archive options in Process Type

ChristianSantamaria:
Adds archive options in Process Type

PR #97, opened 2025-01-15
[13.0][IMP] mrp_unbuild_cust_alu_analytics: add new field start_incidence

Anxo82:
Add new field start_incidence

### DIFF
--- a/mrp_unbuild_cust_alu_analytics/models/mrp_unbuild_incidence.py
+++ b/mrp_unbuild_cust_alu_analytics/models/mrp_unbuild_incidence.py
@@ -6,6 +6,7 @@ from odoo import fields, models
 class MrpUnbuildIncidence(models.Model):
     _name = "mrp.unbuild.incidence"
     _description = 'MRP Incidence'
+    _order = "start_incidence desc"
 
     name = fields.Char(compute='_compute_name')
     detail = fields.Char()
@@ -30,6 +31,9 @@ class MrpUnbuildIncidence(models.Model):
         comodel_name='mrp.incidence.description',
         inverse_name='incidence_type_id',
         string='Description'
+    )
+    start_incidence = fields.Datetime(
+        string='Start Incidence',
     )
 
     def _compute_name(self):
--- a/mrp_unbuild_cust_alu_analytics/views/mrp_unbuild_incidence_views.xml
+++ b/mrp_unbuild_cust_alu_analytics/views/mrp_unbuild_incidence_views.xml
@@ -6,6 +6,7 @@
         <field name="model">mrp.unbuild.incidence</field>
         <field name="arch" type="xml">
             <tree>
+                <field name="start_incidence" />
                 <field name="unbuild_id"/>
                 <field name="unbuild_date"/>
                 <field name="incidence_type_id"/>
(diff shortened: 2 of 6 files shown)


Anxo82:
@ChristianSantamaria @dalonsod Review please.
Thanks

dalonsod:
I'll wait for @ChristianSantamaria review...

ChristianSantamaria:
@Anxo82 Be careful not to overwrite existing translations, you should only translate your new field, nothing else

Anxo82:
@ChristianSantamaria done! review please.

mrp_unbuild_cust_alu_analytics/views/mrp_unbuild_incidence_views.xml
```diff
@@ -11,6 +11,7 @@
                 <field name="incidence_type_id"/>
                 <field name="incidence_description_id"/>
                 <field name="detail"/>
+                <field name="start_incidence" />
```
ChristianSantamaria: I think it is better that this field be the first and that it be ordered by it.

mrp_unbuild_cust_alu_analytics/views/mrp_unbuild_incidence_views.xml
```diff
@@ -25,16 +26,19 @@
             <search>
                 <field name="incidence_type_id"/>
                 <field name="detail"/>
+                <field name="start_incidence"/>
```
ChristianSantamaria: What effect does searching for a Date field in a search view have?

mrp_unbuild_cust_alu_analytics/models/mrp_unbuild_incidence.py
```diff
@@ -32,6 +32,8 @@ class MrpUnbuildIncidence(models.Model):
         string='Description'
     )
 
+    start_incidence = fields.Datetime(string='Start Incidence')
+
```
ChristianSantamaria: Better (TOC)
```suggestion
    start_incidence = fields.Datetime(string='Start Incidence')

```

ChristianSantamaria (changes requested):
Minnor changes

mrp_unbuild_cust_alu_analytics/views/mrp_unbuild_incidence_views.xml
Anxo82: I thought that searching by date would display all incidences for the selected day, but the search actually considers both the date and the time, so it doesn't work as expected.

Anxo82:
@ChristianSantamaria done! changes! review please.

dalonsod (changes requested):
Looks good, but there's a minor issue to be fixed: field default value.

This new field should be required for user, but adding a default value could lead to this situation: incidence start date should be saved in the past (incidence logging should be a deferred task, not real time), user see field already filled and don't update it. We should force user to manually fill them.

Other problem with default value is that past incidences before this improvement will be initialized with installation date, not good...

So I propose this:
- Remove required and default attributes in field definition, so it's open by default.
- For those views that enable incidence editing, add required XML attribute.
- There's a `form` view, we shoud also add the field there, and `required` as well.

Anxo82:
Add field to form view and tree:
imagen
imagen
@dalonsod done! review please.

dalonsod:
@Anxo82 could you make a rebase with `origin/13.0` and fix conflicts due to #96 merge?

@ChristianSantamaria when done, could you install this improvement in AluTest environment?

ChristianSantamaria:
@dalonsod Done

mrp_unbuild_cust_alu_analytics/views/mrp_unbuild_incidence_views.xml
dalonsod: We're all agree with this: search field should be removed

mrp_unbuild_cust_alu_analytics/models/mrp_unbuild_incidence.py
dalonsod: Not agree... I should name field and label as `incidence_start` and "Incidence Start", in proper English IMO... And for this case (or the current one), string attribute is not required, because Odoo will put the desired one without indicating it, at least in this case

dalonsod: Anyway, please don't change anything now, this is a lesson for the next development

mrp_unbuild_cust_alu_analytics/views/mrp_unbuild_incidence_views.xml
```diff
@@ -47,6 +50,7 @@
         <field name="model">mrp.unbuild.incidence</field>
         <field name="arch" type="xml">
             <tree class="tablet_font_size">
```
dalonsod, line 52: Could you indicate sorting by this field here?

mrp_unbuild_cust_alu_analytics/views/mrp_unbuild_incidence_views.xml
```diff
@@ -6,6 +6,7 @@
         <field name="model">mrp.unbuild.incidence</field>
         <field name="arch" type="xml">
             <tree>
+                <field name="start_incidence" />
```
dalonsod, line 9: Could you indicate sorting by this field here?

dalonsod (approved):
LGTM

Only one pending thing: with last rebase bumping manifest version is lost, please add it

PR #101, opened 2025-02-19
[15.0][ADD] mrp_inherit_responsible

IriaAlonso:
Assign the same responsible from a source MO to a child MO, if it has.

When you create a new MO, the responsible is assigned automatically, so then when you assign the MO's products, if it has child, using the `onchange` it doesn't activate, because the responsible doesn't change, so what to do in this case @dalonsod ?


dalonsod:
Yes, if you're right, onchange cannot be used because it doesn't cover all cases. You should try with `create` and `write` instead.

IriaAlonso:
@dalonsod can you review it?

mrp_inherit_responsible/models/mrp_production.py
```diff
@@ -0,0 +1,31 @@
...
+    def write(self, vals):
+        res = super().write(vals)
+        if "user_id" in vals:
+            self._assign_user_id_to_children()
+        return res
+    
+    def action_confirm(self):
+        res = super().action_confirm()
+        self._assign_user_id_to_children()
+        return res
```
dalonsod: Why this action?

dalonsod (approved):
Tested, LGTM

PR #106, opened 2025-03-13
[13.0][ADD] mrp_production_undo + stock_valuation_mrp_undo

dalonsod:
This PR provided two linked addons

**mrp_production_undo**

Enables undo (back to draft) a complete Production, updating stock inventory.

This is enabled only starting from a "done" state and for users that belong to a specific security group. Anyway, this operation can be properly bypassed, if such operation is needed within a server process.

**stock_valuation_mrp_undo**

Links backing to draft a Production with Stock Valuation modules.

With this bridge, valuation layers and related account moves are removed. Then, average prices recomputation in that case is also guaranteed.


dalonsod:
Not strictly approved by customer, but let's merge...

PR #107, opened 2025-03-19
[15.0][FIX] mrp_stock_product_sn_reservation: allow (un)lock S/Ns for "To Close" state

dalonsod:
In "To Close" state some operations, like editing lines quantities, are allowed. Then, it's coherent enabling S/N operations as well.

Before this fix, S/N flag was disabled at such state, so if MO was previously locked, it wasn't possible to revert it. This fixes it.


dalonsod:
:+1: in CtmTest environment, let's merge!

PR #109, opened 2025-03-27
[14.0][IMP] fcd_weight_scale_mrp: some improvements

ChristianSantamaria:
- Parametrizable printer user and pass
- New fields in report tag (origin_country and official_commercial_name)

### DIFF
--- a/fcd_weight_scale_mrp/models/fcd_chkpoint.py
+++ b/fcd_weight_scale_mrp/models/fcd_chkpoint.py
@@ -15,6 +15,8 @@ class FCDCheckpoint(models.Model):
     warehouse_id = fields.Many2one('stock.warehouse')
     printer_ip = fields.Char()
     printer_port = fields.Integer(default=21)
+    printer_user = fields.Char(required=True)
+    printer_pass = fields.Char(required=True)
     external_tara_button = fields.Boolean()
 
     def _default_date(self):
--- a/fcd_weight_scale_mrp/reports/fcd_weight_scale_tag_template.xml
+++ b/fcd_weight_scale_mrp/reports/fcd_weight_scale_tag_template.xml
@@ -2,7 +2,6 @@
 <odoo>
 
     <template id="report_tag_document_pdf">
-      <t t-set="doc" t-value="doc.with_context({'lang':doc.stock_move_id.partner_id.lang})"/>
       <div class="page" style="font-size: 9px; page-break-after: always;">
           <div>
             <div class="row" style="border-bottom: 1px solid black">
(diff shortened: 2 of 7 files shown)


fcd_weight_scale_mrp/models/fcd_chkpoint.py
```diff
@@ -15,6 +15,8 @@ class FCDCheckpoint(models.Model):
     warehouse_id = fields.Many2one('stock.warehouse')
     printer_ip = fields.Char()
     printer_port = fields.Integer(default=21)
+    printer_user = fields.Char()
+    printer_pass = fields.Char()
```
dalonsod: These fields should be required, or `fcd_weight_scale_log.py` comms command should become optional.

dalonsod (changes requested):
See minor comment

dalonsod (approved):
Only code review :+1:

ChristianSantamaria:
imagen

dalonsod:
Translations are missing:

imagen

fcd_weight_scale_mrp/reports/fcd_weight_scale_tag_template.xml
```diff
@@ -73,16 +73,22 @@
...
                           <t t-esc="doc.output_product_id.scientific_name"/>
-                          <span class="d-block">
-                            <span class="font-weight-bold">FAO:</span>
-                            <t t-esc="doc.output_product_id.fao"/>
+                          <span class="row">
+                            <div class="col-6">
+                              <span class="font-weight-bold">FAO:</span>
+                              <span t-field="doc.output_product_id.fao"/>
+                            </div>
+                            <div class="col-6">
```
dalonsod: Make it invisible if country is not filled (as suggested by customer):
```suggestion
                            <div class="col-6" t-if="doc.output_product_id.origin_country_id">
```

fcd_weight_scale_mrp/reports/fcd_weight_scale_tag_template.xml
```diff
@@ -73,16 +73,22 @@
...
+                              <span class="font-weight-bold">ORIGIN COUNTRY:</span>
+                              <span t-field="doc.output_product_id.origin_country_id.name"/>
```
dalonsod: If we directly call `.name`, field shouldn't be translatable. Try this instead:
```suggestion
                              <span t-field="doc.output_product_id.origin_country_id"/>
```

dalonsod (changes requested):
See comments

dalonsod (changes requested):
:+1: 

Only one pending change: squashing last three commits, if I'm right

PR #110, opened 2025-04-16
[13.0][IMP] mrp_unbuild_cust_alu_analytics_valuation: product cost hooks // picking valuation data performance improvement

dalonsod:
Two improvements added:

* Product cost hooks: this is required for individual products cost that comes from Alea
* Picking valuation data performance improvement: picking form load was too load, is moved to a secondary action


dalonsod:
:+1: in AluTest checked along with Alea, let's merge!

PR #111, opened 2025-04-29
[15.0][IMP] mrp_project_analytic: "from child" check new behavior

dalonsod:
Before this commit, existing "analytic_mrp_from_child" check for account move lines was set, when a MO was marked as done, for those stock valuation move lines linked to comsuption lines that actually come from a child production.

After this commit, this check is set for stock valuation account move lines that belong to every child production move itself, not consumption move of the main production.

TODO
- Check in CtmTest environment.
- Update `README`.

HT00660


dalonsod:
@ChristianSantamaria could you apply this update at CtmTest enviroment?

ChristianSantamaria:
Done :+1:

dalonsod:
Tested by LR :+1:

PR #112, opened 2025-05-06
[14.0][FIX] fcd_weight_scale_mrp: specify mrp type in create

ChristianSantamaria:
Specify mrp type in creation, before this modification the default was used


dalonsod (approved):
Only code review :+1:

PR #114, opened 2025-05-12
[15.0][IMP] mrp_project_analytic: "from child" check massive action button

dalonsod:
This action is a techical one, and it should be fired only for those productions that have wrong values for their children, because it unsets the check for children production journal items and, then, set check properly for these children, when required.

E.g. for a production that have children productions and these productions have children itself, it should only affect the two lower levels, leeaving upper level without changes.


dalonsod:
Approved by LR :+1:

PR #116, opened 2025-05-16
[15.0][IMP] mrp_project_analytic_purchase: change function due to improved mrp_project_analytic

ChristianSantamaria:
Change function due to improved mrp_project_analytic

### DIFF
--- a/mrp_project_analytic_purchase/__manifest__.py
+++ b/mrp_project_analytic_purchase/__manifest__.py
@@ -8,10 +8,12 @@
     """,
     "author": "Solvos",
     "license": "LGPL-3",
-    "version": "15.0.1.0.0",
+    "version": "15.0.1.1.0",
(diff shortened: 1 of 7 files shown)


dalonsod:
@ChristianSantamaria rebase made due to #117 and #118

dalonsod (changes requested):
I see this:

* `_get_from_child_acount_move_lines()` is called from `mark_analytic_mrp_from_child()` for a certain productiom.
* So the code that obtain a new set of journal items and set for them the new check `analytic_mrp_from_purchase` could be simply moved from `_get_from_child_acount_move_lines()` to `mark_analytic_mrp_from_child()`.
* With this refactoring, `_get_from_child_acount_move_lines()` will still represent a getter method, and not a getter & updater method.

What do you think @ChristianSantamaria ?

PR #117, opened 2025-05-19
[15.0][IMP] mrp_project_analytic: from child mark, for all journal items

dalonsod:
Before this improvement, only debit journal items could have "from_child" mark. Now, both credit and debit will have it.


dalonsod:
Tested in Ctm Test environment :+1:

PR #118, opened 2025-05-19
[15.0][FIX] mrp_project_analytic: from_child mark smart button form doesn't work sometimes

dalonsod:
Before this, "Set 'From Child' for journal items" form button didn't work when form was loaded from another one (e.g. from project), because of list action management. This fixes it.


dalonsod:
Tested in CtmTest enviroment :+1:

PR #119, opened 2025-06-16
[14.0][IMP] fcd_weight_scale_mrp: change reg_sanitary and commercial_name in QR Tag

ChristianSantamaria:
Change reg_sanitary and commercial_name in QR Tag


dalonsod (approved):
:+1:

PR #129, opened 2025-08-28
[17.0][IMP] mdc_weight_mgmt_xml_import: create new product automatically.

DantePereyra:
**Previous behavior:**
When a product that was not in the database was imported, an error was raised.

**New behavior:**
When a product with a code not present in the database is imported from XML, it will be created automatically with the following fields:

- default_code: code from the XML
- name: name from the XML
- type: 'Consumable'
- categ_id: default category from the company configuration (res.company.product_categ_default_id); if not set, the fallback will be 'All'

Also, in product, a log message it's displayed
imagen
And log WARNING
imagen

### DIFF
--- a/mdc_weight_mgmt_xml_import/models/maintenance_equipment.py
+++ b/mdc_weight_mgmt_xml_import/models/maintenance_equipment.py
@@ -196,6 +196,7 @@ class MaintenanceEquipment(models.Model):
         root = tree.getroot()
         nsmap = root.nsmap
 
+        product_name = self.get_statvalue_by_id('102', root=root, nsmap=nsmap)
         product_code = self.get_statvalue_by_id('103', root=root, nsmap=nsmap)
         machine_text = self.get_text_from('Machine_Number', root=root, nsmap=nsmap)
         start_str = self.get_text_from('Production_Start', root=root, nsmap=nsmap)
@@ -225,8 +226,31 @@ class MaintenanceEquipment(models.Model):
                     found=equipment.name if equipment else "None"
                 ))
             product = self.env['product.product'].search([('default_code', '=', product_code)], limit=1)
-            if not product or not product_code:
-                raise Exception(_("Product with code %(code)s not found", code=product_code))
+            if not product:
+                categ = (
+                    record.company_id.product_categ_default_id
+                    or self.env.ref("product.product_category_all", raise_if_not_found=False)
+                    or self.env['product.category'].search(
+                        [('company_id', '=', record.company_id.id)] if record.company_id else [('company_id', '=', False)],
+                        limit=1
+                    )
+                )
+                product_template = self.env['product.template'].sudo().create({
+                    'name': product_name,
+                    'default_code': product_code,
+                    'detailed_type': 'consu',
+                    'categ_id': categ.id,
+                    'company_id': record.company_id.id or False,
+                })
+                product = product_template.product_variant_id
+                logger.warning(
+                    "Product with code %s not found. Automatically created with name '%s'.",
+                    product_code,product_name
+                )
+                product.message_post(
+                    body=_("Automatically created from file '%(file)s'",file=file.name),
+                    subtype_id=self.env.ref("mail.mt_comment").id
+                )
 
             msg = self.env['mdc.weight.record']._check_overlap_manual(equipment, start_dt, end_dt)
             if msg:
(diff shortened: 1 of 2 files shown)


mdc_weight_mgmt_xml_import/models/maintenance_equipment.py
```diff
@@ -223,8 +224,28 @@ def _process_xml_file(self, file, processed_path, record):
                 found=equipment.name if equipment else "None"
             ))
         product = self.env['product.product'].search([('default_code', '=', product_code)], limit=1)
-        if not product or not product_code:
-            raise Exception(_("Product with code %(code)s not found", code=product_code))
+        if not product:
+            categ = getattr(record.company_id, 'product_categ_default_id', None)
```
dalonsod: Please explain this line, I'm not sure why `getattr()` is used.

mdc_weight_mgmt_xml_import/models/maintenance_equipment.py
```diff
@@ -223,8 +224,28 @@ def _process_xml_file(self, file, processed_path, record):
...
+            if not categ:
+                categ = self.env.ref("product.product_category_all")
```
dalonsod: Good point! But if `All` category was removed, this will raise an error. Check `raise_if_not_found` parameter. For this very corner case, we could obtain the first category found for equipment company (ir without company, if equipment has no company assigned)

mdc_weight_mgmt_xml_import/models/maintenance_equipment.py
```diff
@@ -223,8 +224,28 @@ def _process_xml_file(self, file, processed_path, record):
...
+
+            product_template = self.env['product.template'].create({
```
dalonsod: There's a hidden parameter that should be filled (or not): company_id. I've decided to be filled with equipment company, if filled, to be coherent.

mdc_weight_mgmt_xml_import/models/maintenance_equipment.py
```diff
@@ -223,8 +224,28 @@ def _process_xml_file(self, file, processed_path, record):
...
+            })
+            product = product_template.product_variant_id
+
+            logger.warning(
+                "Product with code %s not found. Automatically created with name '%s'.",
+                product_code,product_name
+            )
+
+            product.message_post(
+                body=_("Automatically created from file '%s'",file.name),
```
dalonsod: Could you use new named parameter management for `_` method call?

mdc_weight_mgmt_xml_import/models/maintenance_equipment.py
dalonsod: Please ensure that product is created with `sudo()`, because we don't know if cron user has the right permissions

dalonsod (changes requested):
See comments. And remember update module version at manifest

mdc_weight_mgmt_xml_import/models/maintenance_equipment.py
DantePereyra: Originally considered using `getattr()` for safety in case the field was missing,
but since this module depends on the addon that defines it, maybe accessing
`record.company_id.product_categ_default_id` directly is cleaner and sufficient.

mdc_weight_mgmt_xml_import/models/maintenance_equipment.py
DantePereyra: Now, in order checks this way:
- First, it uses the company’s default category.
- If it does not exist, it falls back to the `All` category if available.
- If that is also missing, it falls back to the first category available for the equipment’s company (or without a company if the equipment has no company assigned).

Let me know if this is the expected behavior.

mdc_weight_mgmt_xml_import/models/maintenance_equipment.py
dalonsod: Well, if `record.company_id.product_categ_default_id` does not raise an error (and return e.g. `False`) when `record.company_id` is `False` as well, it will work. Could you check?

DantePereyra: We would only raise an error if `product_categ_default_id` does not exist. However, this field is provided by the `mdc_weight_mgmt` addon, which is a dependency of this module, so this situation should never occur. Also works when `record.company_id` is `False`

dalonsod (approved):
LGTM, let's go!

PR #130, opened 2025-09-03
[17.0][IMP] mdc_weight_mgmt_xml_import: Skip creating weight records for em…

DantePereyra:
…pty XML files

Files where both accepted and rejected units are 0 are now ignored. A warning is logged and the file is moved to the processed folder.


dalonsod (approved):
:+1: at MsvTest environment, let's merge!

PR #132, opened 2025-09-12
[17.0][IMP] mdc_weight_mgmt_xml_import: Automatic deletion of processed files

DantePereyra:
With this imp, you can set up an automatic deletion of processed XML files after a certain number of days. This is configured in the Maintenance Equipment form view, where a new field "Days to Delete Processed Files" has been added. If this field is set to a value greater than zero, a scheduled action will run daily to delete files in the processed path that are older than the specified number of days.

### DIFF
--- a/mdc_weight_mgmt_xml_import/data/ir_cron_mdc_weight_import.xml
+++ b/mdc_weight_mgmt_xml_import/data/ir_cron_mdc_weight_import.xml
@@ -9,4 +9,16 @@
         <field name="state">code</field>
         <field name="code">model._cron_import_xml_weight()</field>
     </record>
+
+    <record id="ir_cron_mdc_cleanup_processed_files" model="ir.cron">
+        <field name="name">Automated clean processed XML Weight Files</field>
+        <field name="user_id" ref="base.user_root" />
+        <field name="active" eval="False" />
+        <field name="interval_number">1</field>
+        <field name="interval_type">days</field>
+        <field name="numbercall">-1</field>
+        <field name="model_id" ref="model_maintenance_equipment" />
+        <field name="state">code</field>
+        <field name="code">model._cron_mdc_xml_cleanup_processed_files()</field>
+    </record>
 </odoo>
--- a/mdc_weight_mgmt_xml_import/models/maintenance_equipment.py
+++ b/mdc_weight_mgmt_xml_import/models/maintenance_equipment.py
@@ -27,6 +27,18 @@ class MaintenanceEquipment(models.Model):
         string="XML Processed Folder Path",
         help="Path to the folder where processed XML files are moved."
     )
+    mdc_timezone = fields.Selection(
+        selection=_tz_get,
+        string="Timezone",
+        default=lambda self: self.env.user.tz or 'UTC',
+        help="Timezone for the equipment."
+    )
+    mdc_xml_days_delete = fields.Integer(
+        string="Days to Keep Processed Files",
+        default=90,
+        help="""Delete files older than this number of days from the processed folder.
+        If set to less than 1, files will be kept indefinitely.""",
+    )
     # FTP connection settings
     requires_ftp = fields.Boolean(
         string="Requires FTP Connection",
@@ -37,13 +49,6 @@ class MaintenanceEquipment(models.Model):
     ftp_user = fields.Char(string="Username")
     ftp_password = fields.Char(string="Password")
     ftp_folder = fields.Char(string="Folder")
-
-    mdc_timezone = fields.Selection(
-        selection=_tz_get,
-        string="Timezone",
-        default=lambda self: self.env.user.tz or 'UTC',
-        help="Timezone for the equipment."
-    )
 
     @api.model
     def _cron_import_xml_weight(self):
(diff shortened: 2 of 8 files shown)


mdc_weight_mgmt_xml_import/data/ir_cron_mdc_weight_import.xml
```diff
@@ -1,12 +1,25 @@
-<odoo noupdate="1">
+<odoo noupdate="0">
```
dalonsod: This seems to be wrong, why this change?

mdc_weight_mgmt_xml_import/models/maintenance_equipment.py
```diff
@@ -45,6 +45,12 @@ class MaintenanceEquipment(models.Model):
         help="Timezone for the equipment."
     )
 
+    mdc_xml_days_delete = fields.Integer(
+        string="Days to Keep Processed Files",
+        default=90,
+        help="Delete files older than this number of days from the processed folder."
```
dalonsod: Improve a little bit this, adding that if number of days is less than 1 (.e.g. 0) deletion process will be skipped

mdc_weight_mgmt_xml_import/models/maintenance_equipment.py
```diff
@@ -118,6 +124,36 @@ def _cron_import_xml_weight(self):
             error_model = self.env['mdc.weight.record.error'].sudo()
             created_errors = error_model.create(errors)
             created_errors.action_send_mail()
+
+    @api.model
+    def _cron_cleanup_processed_files(self):
```
dalonsod: Rename this method and make it more specific (e.g. adding `mdc_xml` in name)

mdc_weight_mgmt_xml_import/models/maintenance_equipment.py
```diff
@@ -118,6 +124,36 @@ def _cron_import_xml_weight(self):
...
+                    if age_days > record.mdc_xml_days_delete:
+                        try:
+                            file.unlink()
+                            deleted_files.append(file.name)
+                        except Exception as e:
+                            logger.error("Error deleting file %s: %s", file.name, e)
+
+            if deleted_files:
+                logger.warning(
+                    "Deleted processed XML files for equipment %s: %s",
```
dalonsod: Add `record.mdc_xml_days_delete` content to log message in some way

dalonsod (changes requested):
Only code review, see comments

mdc_weight_mgmt_xml_import/data/ir_cron_mdc_weight_import.xml
DantePereyra: My bad, was for a local use.

mdc_weight_mgmt_xml_import/data/ir_cron_mdc_weight_import.xml
```diff
@@ -5,8 +5,21 @@
         <field name="active" eval="True" />
         <field name="interval_number">1</field>
         <field name="interval_type">hours</field>
+        <field name="numbercall">-1</field>
```
dalonsod: Remove this, it could be later added in a different PR (that will need a `noupdate_changes.xml` file, anyway), not planned right now

mdc_weight_mgmt_xml_import/models/maintenance_equipment.py
```diff
@@ -45,6 +45,13 @@ class MaintenanceEquipment(models.Model):
         help="Timezone for the equipment."
     )
 
+    mdc_xml_days_delete = fields.Integer(
+        string="Days to Keep Processed Files",
+        default=90,
+        help="""Delete files older than this number of days from the processed folder.
+        If set to less than 1, files will be kept indefinitely.""",
+    )
+
```
dalonsod, line 52: Could you move these lines closer to the other `mdc_xml_*` fields?

mdc_weight_mgmt_xml_import/views/maintenance_equipment_views.xml
```diff
@@ -26,6 +26,7 @@
                                 name="mdc_xml_weight_processed_path"
                                 required="mdc_xml_import_active == True"
                             />
+                            <field name="mdc_xml_days_delete"/>
```
dalonsod, line 29: Could you provide a screenshot for this field location?

DantePereyra: imagen

dalonsod:
Good, could you install this in McdTest environment? Before installing it, first ensure that cron is deactivated. Once done, we'll test it together.

DantePereyra:
Done 
imagen

dalonsod (approved):
:+1: in MsvTest enviroment.

Please add pending `es` translations.

DantePereyra:
@dalonsod, check if file translations are ok like this,  because it brings some changes that had previous IMP that was mentioned here #131. I could avoided but I'm not sure if it will be a problem.

dalonsod:
> @dalonsod, check if file translations are ok like this, because it brings some changes that had previous IMP that was mentioned here #131. I could avoided but I'm not sure if it will be a problem.

Merged #131 , you can rebase this and update

PR #136, opened 2025-09-23
[17.0][IMP] mdc_weight_mgmt(_xlsx_report): make shift optional and product multiple selection.

DantePereyra:
Before this IMP's, shift was required and only can select one product, if none was selected it showed all.
Now shift it's optional and can select more than one product.

imagen

SlvTicket: HT00922


dalonsod (approved):
:+1: 

Only one thing to be reviewed:

imagen

This legend should be displayed in plural now.

PR #137, opened 2025-09-23
[17.0][FIX] mdc_weight_mgmt_xlsx_report: Adjust XLSX report to use user tim…

DantePereyra:
…ezone

Previously XLSX report was showing start and end times in UTC timezone, now it shows them in the user's timezone as displayed in the Odoo interface.


dalonsod (approved):
LGTM

PR #140, opened 2025-12-03
[17.0][FIX/UPD] mdc_weight_mgmt: overlap function and declared weight field

DantePereyra:
First commit attends an issue that when creating a weight record manually, funtion `_check_overlap_constrains()`
was raising an error even when there was no overlapping. The problem was that the record was being created before the constrains were checked, so the search was returning the same record being created, causing a false positive. Now it's checked on `create()`

Second commit, before this update, the field `weight_dec_qty` in the `mdc_weight_record` model was input manually by record. This update changes it to be computed from the product's `weight_dec_qty` field, ensuring consistency and reducing manual errors.

SLV-Ticket: HT01037


mdc_weight_mgmt/models/product_product.py
```diff
@@ -0,0 +1,25 @@
...
+class ProductProduct(models.Model):
+    _inherit = 'product.product'
+
+    def _get_default_category_id(self):
+        context = self.env.context
+        if context.get('product_categ_default_id'):
+            return self.env.company.product_categ_default_id
+        return super(ProductProduct,self)._get_default_category_id()
+
+    weight_dec_qty = fields.Float(
```
dalonsod: Prefix this, as `maintenance.equipment` model fields do.

mdc_weight_mgmt/models/product_product.py
```diff
@@ -0,0 +1,25 @@
...
+        string='Declared Weight',
+        required=True
+    )
+
+    @api.constrains("weight_dec_qty")
+    def _check_weight_dec_qty(self):
+        for product in self:
+            if product.weight_dec_qty <= 0:
+                raise ValidationError(_("The declared weight must be greater than 0."))
```
dalonsod: Suggested code (prevent `for` + `if` usage), not tested:
```suggestion
        products_ko_weight_dec_qty = self.filtered(lambda x: x.weight_dec_qty <= 0)
        if products_ko_weight_dec_qty:
            raise ValidationError(_("The declared weight must be greater than 0."))
```

mdc_weight_mgmt/models/product_template.py
```diff
@@ -11,3 +12,22 @@ def _get_default_category_id(self):
         if context.get('product_categ_default_id'):
             return self.env.company.product_categ_default_id
         return super(ProductTemplate,self)._get_default_category_id()
+
+    weight_dec_qty = fields.Float(
+        string="Declared Weight",
+        compute="_compute_weight_dec_qty",
+        inverse="_inverse_weight_dec_qty",
+        store=True,
```
dalonsod: This could be removed
```suggestion
```

mdc_weight_mgmt/models/product_template.py
```diff
@@ -11,3 +12,22 @@ def _get_default_category_id(self):
...
+    @api.depends("product_variant_ids.weight_dec_qty")
+    def _compute_weight_dec_qty(self):
+        for template in self.filtered('product_variant_ids'):
+            template.weight_dec_qty = template.product_variant_ids[0].weight_dec_qty
+
+    def _inverse_weight_dec_qty(self):
+        for template in self.filtered('product_variant_ids'):
+            value = template.weight_dec_qty
+            if value <= 0:
+                raise ValidationError(_("The declared weight must be greater than 0."))
```
dalonsod: Is this check necessary, existing constrains for `product.product` field?

mdc_weight_mgmt/models/product_product.py
```diff
@@ -0,0 +1,25 @@
+# © 2025 Solvos Consultoría Informática (<http://www.solvos.es>)
+# License LGPL-3 - See http://www.gnu.org/licenses/lgpl-3.0.html
+from odoo import models, fields, api, _
+from odoo.exceptions import ValidationError
+
+
+class ProductProduct(models.Model):
+    _inherit = 'product.product'
+
+    def _get_default_category_id(self):
```
dalonsod: This method seems to be unnecessary

mdc_weight_mgmt/models/mdc_weight_record.py
```diff
@@ -143,29 +153,19 @@ def write(self, vals):
...
-                continue
-            msg = self._check_overlap_manual(record.equipment_id, record.start, record.end)
-            if msg:
-                raise ValidationError(msg)
-
-    @api.model
     def _check_overlap_manual(self, equipment_id, start, end):
         overlapping = self.search([
-            ('equipment_id', '=', equipment_id.id),
+            ('equipment_id', '=', equipment_id),
```
dalonsod: I'll give you an example for `equipment_id` possible dual value (between record and int value)

dalonsod (changes requested):
See comments

mdc_weight_mgmt/models/mdc_weight_record.py
dalonsod: The example (key word: `isinstance`)

https://github.com/odoo/odoo/blob/9540359d6a2efd96304f4ff7f22d48661255af71/addons/stock/models/product.py#L280-L281

For our case yo could use `isinstance` with `int` to check when `id` value is passed

dalonsod (approved):
:+1:

PR #141, opened 2025-12-09
[17.0][IMP] mdc_weight_mgmt_xml_import: added neutralize stuff

dalonsod:


dalonsod:
@ChristianSantamaria @DantePereyra you can check here how to configure a neutralize script for an Odoo module. With this, when importing via `click-odoo` a database with `--neutralize` option (only available for `16.0` and above), that script will be executed.

dalonsod:
:+1:  at MscTest environment

PR #143, opened 2026-01-02
[17.0][FIX] mdc_weight_mgmt: Fix division by zero in calculations

DantePereyra:
When calculating average, percentage average and total nominal weight in the `weight_record_report_wizard`, ensure that division by zero is handled gracefully by returning 0.0 when the denominator is zero. This prevents runtime errors and ensures the stability of the module during report generation.

SLV-Ticket: HT01062

PR #144, opened 2026-01-05
[17.0][FIX] mdc_weight_mgmt[_xlsx_report]: Correct total nominal average calculation

DantePereyra:
Previously, calculations followed the provided example without including the nominal value in the product total, as it was not expected to change. This has now been updated to calculate and internally use the nominal value to ensure the average is computed correctly according to the original logic.

Manifest remains untouched until #143 is merged, so it can be tested in the MSC_Test environment.

SLV-TICKET: HT01063

### DIFF
--- a/mdc_weight_mgmt/__manifest__.py
+++ b/mdc_weight_mgmt/__manifest__.py
@@ -8,7 +8,7 @@
     """,
     "author": "Solvos",
     "license": "LGPL-3",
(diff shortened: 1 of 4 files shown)


dalonsod (approved):
:+1:  I've slightly updated PR title, according to two different addons fixed. Let's merge...

PR #145, opened 2026-01-08
[15.0][IMP] mrp_unbuild_mo_repair: remove move with qty < 0

ChristianSantamaria:
Remove move with qty < 0


dalonsod:
You can install this at CtmTest and we could test it together

PR #146, opened 2026-01-12
[17.0][UPD] mdc_weight_mgmt: report group by hour

DantePereyra:
As mdc weight records are now imported more frequently, it is useful to have the report grouped by hour to have a better overview of the data. This commit implements that change by creating a record on the fly for each hour with the aggregated data of that hour. If data ends before the hour ends, the end time is set accordingly.

Before:
imagen
After:
imagen

SLV-Ticket: HT01081


dalonsod:
:+1: could you install this at MscTest environment?

dalonsod (approved):
LGTM :+1:

PR #149, opened 2026-01-20
[15.0][FIX] mrp_project_wo_timesheet: add recalculate timesheet in create function

ChristianSantamaria:
Add recalculate timesheet in create function

### DIFF
--- a/mrp_project_wo_timesheet/models/mrp_workcenter_productivity.py
+++ b/mrp_project_wo_timesheet/models/mrp_workcenter_productivity.py
@@ -1,7 +1,7 @@
 # © 2025 Solvos Consultoría Informática (<http://www.solvos.es>)
 # License LGPL-3 - See https://www.gnu.org/licenses/lgpl-3.0.html
 
-from odoo import fields, models, _
+from odoo import fields, models, api, _
 from odoo.exceptions import UserError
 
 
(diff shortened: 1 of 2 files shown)


mrp_project_wo_timesheet/models/mrp_workcenter_productivity.py
```diff
@@ -10,13 +10,25 @@ class MRPWorkcenterProductivity(models.Model):
 
     timesheet_id = fields.Many2one('account.analytic.line', copy=False)
 
+    def _check_production_project_id(self):
+        if not self.filtered(lambda x: x.production_id.project_id):
```
dalonsod: This condition returns `true` if one of the productivities has a related `project_id`, then, a productivity without project could pass this (and this should be wrong). Not tested, try this:
```suggestion
        if not all([project_id for project_id in self.production_id]):
```

mrp_project_wo_timesheet/models/mrp_workcenter_productivity.py
```diff
@@ -10,13 +10,25 @@ class MRPWorkcenterProductivity(models.Model):
...
+            raise UserError(_('To generate the related timesheets, you need to specify the project for this manufacturing.'))
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        mwp_ids = super(MRPWorkcenterProductivity, self).create(vals_list)
+        for record in mwp_ids.filtered(lambda x: x.date_end):
+            self._check_production_project_id()
```
dalonsod: Aghhh the monster!!!! `self` used within `for` statement:
```suggestion
            record._check_production_project_id()
```

dalonsod (changes requested):
See minor comments, it should cover special multi cases not properly covered now IMO.

Once updated, we could merge this.

dalonsod (approved):
:+1:

PR #150, opened 2026-01-28
[17.0][FIX] mdc_weight_mgmt_xlsx_report: correct total units per minute calculation

DantePereyra:
Previously, the average was computed using only the accepted units, which led to inaccurate reporting. Total product units per minute are now calculated correctly in the XLSX report using total units.


dalonsod (approved):
:+1:

PR #152, opened 2026-02-24
[17.0][FIX] mdc_weight_mgmt: Fix unit_x_min calculation in report wizard

DantePereyra:
PR https://github.com/solvosci/slv-manufacture/pull/146 added report grouping by hour. When creating record on the fly, the calculation of unit_x_min was using unit_ok instead of unit_total, which could lead to incorrect results when there are rejected units. This commit corrects the calculation to use unit_total, ensuring that the average is calculated based on the total number of units.


dalonsod (approved):
👍

PR #153, opened 2026-02-25
[17.0][IMP] mdc_weight_mgmt: Add daily report for weight records

DantePereyra:
With this IMP, we add a direct access to a daily report for weight records, which can be accessed from the menu that opens the report directly with the current day as filter.

SLV-Ticket: HT01154
@dalonsod


dalonsod:
Great! Could you install it at MscTest Environment?

DantePereyra:
Done!

dalonsod (approved):
👍  at MscTest environment. Let's merge!

PR #157, opened 2026-04-01
[15.0][ADD] mrp_account_val_operation: Add valuation operation for MRP processes

DantePereyra:
This addon extends the functionality of `stock_account_val_operation` to include valuation operations for MRP processes. Now manufacturing orders and unbuilds will also have valuation operations for the produced and consumed items.

Depends on [stock_account_val_operation](https://github.com/solvosci/slv-stock/pull/215)
SLV-Ticket: HT01144


mrp_account_val_operation/models/account_move.py
```diff
@@ -0,0 +1,43 @@
...
+        ('mrp_cons', 'Manufacture Consumption'),
+        ('mrp_prod', 'Manufacture Production'),
+        ('ub_cons', 'Disassembly - Re-entry of consumed items'),
+        ('ub_prod', 'Disassembly - Previously produced items')],
+    )
+    #TODO: Add ondelete and handle the unlink of those records in case addon is uninstalled
+
+    def _compute_val_operation(self):
+        super()._compute_val_operation()
+        lines_with_move = self.filtered(lambda l: l.move_id.stock_move_id)
```
dalonsod: Result of filtered can be reduced adding this:

```suggestion
        lines_with_move = self.filtered(lambda l: l.move_id.stock_move_id and (l.raw_material_production_id or l.production_id or l.unbuild_id))
```

mrp_account_val_operation/hooks.py
```diff
@@ -0,0 +1,28 @@
+# © 2026 Solvos Consultoría Informática (<http://www.solvos.es>)
+# License LGPL-3.0 (https://www.gnu.org/licenses/lgpl-3.0.html)
+import logging
+
+_logger = logging.getLogger(__name__)
+
+def pre_init_hook(cr):
+    cr.execute("""
+        UPDATE account_move_line aml
```
dalonsod, line 9: This hook is not actually covering those unbuilds that don't come from an MO. We'll not cover this yet (it's actually a corner case for our customer), but add it to `ROADMAP`

dalonsod (changes requested):
See minor comments

dalonsod (approved):
👍  and for @lromero-solvos too

PR #159, opened 2026-04-23
[15.0][FIX] mrp_project: assign parent MO project to child MO

IriaAlonso:
When a project is modified in a MO that has child MOs, the changes are automatically applied to those child MOs as well.

This only happens if the child MOs already exist;
if they are generated later, the changes are not applied. This fix resolves that issue.
@dalonsod

### DIFF
--- a/mrp_project/models/mrp_production.py
+++ b/mrp_project/models/mrp_production.py
@@ -26,3 +26,12 @@ class ManufactureOrder(models.Model):
     @api.onchange("project_id")
     def _onchange_project_id_parent(self):
         self._assign_project_to_children()
+
+    @api.model
+    def create(self, values):
+        # This part of the code is triggered by another automated process and is executed by OdooBot with permissions
+        production = super().create(values)
+        production_source = production._get_sources()
+        if production_source and len(production_source) == 1:
+            production.project_id = production_source.project_id
+        return production
--- a/mrp_project/__manifest__.py
+++ b/mrp_project/__manifest__.py
@@ -7,7 +7,7 @@
     """,
     "author": "Solvos",
     "license": "LGPL-3",
-    "version": "15.0.2.0.0",
+    "version": "15.0.2.0.1",
     "category": "Manufacturing",
     "website": "https://github.com/solvosci/slv-manufacture",
     "depends": ["mrp", "project"],


dalonsod:
Thanks @IriaAlonso . I'll wait for @DantePereyra 's review...

mrp_project/models/mrp_production.py
```diff
@@ -26,3 +26,10 @@ def _assign_project_to_children(self):
     @api.onchange("project_id")
     def _onchange_project_id_parent(self):
         self._assign_project_to_children()
+
+    @api.model
```
DantePereyra, line 30: This shouldn't be `model_create_multi`? MO's child could be more than one.

mrp_project/models/mrp_production.py
```diff
@@ -26,3 +26,10 @@ def _assign_project_to_children(self):
...
+    def create(self, values):
+        production = super().create(values)
+        if production.origin:
+            production.project_id = self.env['mrp.production'].search([("name", "=", production.origin)]).project_id
+        return production
```
DantePereyra, line 37:
```suggestion
    @api.model_create_multi
    def create(self, vals_list):
        for vals in vals_list:
            origin = vals.get('origin')
            if origin:
                vals['project_id'] = self.env['mrp.production'].search([("name", "=", origin)]).project_id.id
        return super().create(vals_list)
```

Could you try this if it gets same result?

DantePereyra (changes requested):
Some minor suggestions

mrp_project/models/mrp_production.py
dalonsod: @DantePereyra good point. But, in this case, as original addon actually uses basic `create` option, at least at `15.0` (see https://github.com/odoo/odoo/blob/3a28e5b0adbb36bdb1155a6854cdfbe4e7f9b187/addons/mrp/models/mrp_production.py#L796 ), it's better adapt to it IMO.

mrp_project/models/mrp_production.py
IriaAlonso:
```suggestion
    @api.model
    def create(self, vals_list):
        origin = vals_list.get('origin')
        if origin:
            vals_list['project_id'] = self.env['mrp.production'].search([("name", "=", origin)]).project_id.id
        return super().create(vals_list)
```
In this way, we would get the same result

mrp_project/models/mrp_production.py
```diff
@@ -26,3 +26,11 @@ def _assign_project_to_children(self):
     @api.onchange("project_id")
     def _onchange_project_id_parent(self):
         self._assign_project_to_children()
+
+    @api.model
+    def create(self, values):
+        # This part of the code is triggered by another automated process and is executed by OdooBot with permissions
+        production = super().create(values)
+        if production._get_sources():
+            production.project_id = production._get_sources().project_id
```
dalonsod: This code should be properly protected. As we know, `_get_sources()` could return more than one MO, then `production._get_sources().project_id` could return a recordset with two or more project, an this code could raise an error. Better try this (not tested, we inspect of only one project is available before assigning it):

```suggestion
        project_ids = production._get_sources().project_id
        production.project_id = (
          len(project_ids) == 1
          and project_ids
          or False
        )
```

dalonsod (approved):
👍  ready to be merged!

PR #161, opened 2026-05-14
[15.0][ADD] quality_control_stock_oca_picking_context

DantePereyra:
Add new module for picking context in quality control stock OCA

SLV-Ticket: HT01255


dalonsod (approved):
LGTM. Before merging this, please add a `ROADMAP` indicating that it's intended to add this to main `quality_control_stock_oca` addon as an improvement, and once done, this addon will become unnecessary.

DantePereyra:
@dalonsod done!

dalonsod (commented):
Making a functiona review, I've seen this: without this addon installed (or creating an inspection from scratch), if we want to link an inspection to a certain picking, we need to first properly fill the `Reference` field` reference`:

Image

With this addon installed, creating a test from an existing picking form, `picking` is properly filled (OK), but `reference` field not:

Image

Could you check if this is a potential problem, or if `reference` field is not actually relevant and we could skip this?

dalonsod:
@DantePereyra added to `ROADMAP` pending feature (filling `reference` field). Let's merge...

PR #164, opened 2026-06-01
[15.0][FIX] quality_control_oca_report: Remove internal notes from report

DantePereyra:
`internal_notes` were added as comments in the inspection report, this commit removes it from the report template.
`external_notes`, are already added on template.
 
SLV-Ticket: HT01284
@dalonsod

### DIFF
--- a/quality_control_oca_report/__manifest__.py
+++ b/quality_control_oca_report/__manifest__.py
@@ -7,7 +7,7 @@
     """,
     "author": "Solvos",
     "license": "AGPL-3",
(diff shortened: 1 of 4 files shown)


dalonsod (approved):
LGTM

Due to this fix that removes code, you should update `po` and `pot` files, because there should be at least one key entry that should be removed. Once done, I'll merge it

DantePereyra:
@dalonsod, change done! Added `.po` and `.pot`

PR #167, opened 2026-06-18
[15.0][ADD] mrp_production_label_customization

DantePereyra:
This addon changes strings of `date_planned_start` and `date_deadline` fields, and also removes `widget='remaining_days'` on `tree view` for better clarity.

This comes from SLV-Ticket :HT01251 as MOD 0


mrp_production_label_customization/views/mrp_production_view.xml
```diff
@@ -0,0 +1,35 @@
...
+            <label for="date_planned_start" position="attributes">
+                <attribute name="string">Schedule Start Date</attribute>
+            </label>
+        </field>
+    </record>
+
+    <record id="mrp_production_tree_view" model="ir.ui.view">
+        <field name="name">mrp.production.tree</field>
+        <field name="model">mrp.production</field>
+        <field name="inherit_id" ref="mrp.mrp_production_tree_view"/>
```
dalonsod: As `date_planned_finished` isn't actually available for this inherited from view, it should be replaced by the proper one:

```suggestion
        <field name="inherit_id" ref="mrp_production_date_planned_finished.mrp_production_tree_view"/>
```

dalonsod (approved):
Not totally agree with addon name (due to its dependency), but it's not very important. Let's go...

PR #171, opened 2026-07-09
[14.0][IMP] fcd_weight_scale_mrp: add quantity logs in packaging system

ChristianSantamaria:
add quantity logs in packaging system
Need create parameter system:
fcd_weight_scale_mrp.enable_lot_logs = True


dalonsod (approved):
Code & functionally tested 👍 

Let's merge!

PR #173, opened 2026-07-13
[13.0][FIX] mrp_unbuild_cust_alu_analytics: proper BoM domain and default selection with process type restriction addition

dalonsod:
Before this fix an incomplete domain was set for unbuild bill of materials when process type was set:
* Comparing to standard default, company and Bom type conditions were removed
* Variant condition was removed as well, resulting in every template product BoM available.

Additionally, default Bom value selection when product and process type
are changed is fixed as well, because it wasn't taking in account
product variant management.

This commit fixes these problems together.


dalonsod:
Sucessfully tested at customer Test environment, let's merge!

PR #174, opened 2026-07-14
[19.0][ADD] mrp_production_classification, mrp_operation_type_dashboard and mrp_production_classification_stock_lot_name_sequence and stock_lot_name_sequence

cgarcia-solvos:


mrp_production_classification/models/mrp_production.py
```diff
@@ -0,0 +1,77 @@
...
+from odoo.exceptions import UserError
+
+
+class MrpProduction(models.Model):
+    _inherit = "mrp.production"
+
+    is_classification_type = fields.Boolean(
+        related="picking_type_id.is_classification",
+        string="Is Classification",
+        store=False,
```
dalonsod, line 14: This is unnecessary, as default condition

mrp_production_classification/models/mrp_production.py
```diff
@@ -0,0 +1,77 @@
...
+        "time the classification wizard was opened.",
+    )
+
+    def action_open_classification_wizard(self):
+        self.ensure_one()
+
+        raw_moves = self.move_raw_ids.filtered(lambda m: m.state != "cancel")
+        if len(raw_moves) != 1:
+            raise UserError(
+                _(
```
dalonsod: Starting from `18.0`, this is recommended instead of `_` function usage, change here an for the other usages:

```suggestion
                self.env._(
```

mrp_production_classification/models/mrp_production.py
```diff
@@ -0,0 +1,77 @@
...
+            "name": _("Classification Result"),
+            "res_model": "mrp.classification.wizard",
+            "res_id": wizard.id,
+            "view_mode": "form",
+            "target": "new",
+        }
+
+    def button_mark_done(self):
+        classification_orders = self.filtered("is_classification_type")
+        if self - classification_orders:
```
dalonsod, line 72: I don't understand this. If this is multi, why not obtain two different sets and process them separately, depending on classification type? This code seems to indicate that "all of them are classifications, or no one", and that doesn't seem to be OK

mrp_production_classification/wizard/mrp_classification_wizard.py
```diff
@@ -0,0 +1,275 @@
...
+
+    production_id = fields.Many2one("mrp.production", required=True, readonly=True)
+    raw_move_id = fields.Many2one(
+        "stock.move",
+        required=True,
+        readonly=True,
+        help="Move of the bulk component being consumed.",
+    )
+    original_product_id = fields.Many2one(
+        related="raw_move_id.product_id", string="Original Product", readonly=True
```
dalonsod: I think that attribute is unrequired, because that is the default value:

```suggestion
        related="raw_move_id.product_id", string="Original Product"
```

mrp_production_classification/wizard/mrp_classification_wizard.py
```diff
@@ -0,0 +1,275 @@
...
+    )
+
+    qty_to_consume = fields.Float(
+        string="Planned Quantity (Initial Demand)",
+        digits="Product Unit",
+        readonly=True,
+    )
```
dalonsod, line 31: I miss here product unit, to be shown at `form` view.

mrp_production_classification/__manifest__.py
```diff
@@ -0,0 +1,19 @@
...
+{
+    "name": "MRP Production Clasification",
+    "summary": "Adds a Wizard to distribute quantities and lots for 'Classification' "
+    "type manufacturing orders (bulk product -> N sizes/byproducts)",
+    "author": "Solvos, Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/manufacture",
+    "category": "Manufacturing",
+    "version": "19.0.1.0.0",
+    "license": "LGPL-3",
+    "depends": ["mrp", "stock"],
```
dalonsod: `mrp` actually depends on `stock`, so it can be safely removed

```suggestion
    "depends": ["mrp"],
```

stock_lot_name_sequence/models/stock_lot.py
```diff
@@ -0,0 +1,81 @@
...
+from odoo import api, fields, models
+
+
+class StockLot(models.Model):
+    _inherit = "stock.lot"
+
+    @api.onchange("product_id")
+    def _onchange_product_id_suggest_name(self):
+        for lot in self:
+            if lot.product_id:
```
dalonsod: @cgarcia-solvos you know how should this be actualy coded... 😇

stock_lot_name_sequence/readme/DESCRIPTION.md
```diff
@@ -0,0 +1 @@
+Adds a Wizard to distribute quantities and lots for 'Classification' type manufacturing orders (bulk product -> N sizes/byproducts)
```
dalonsod: Wrong description, probably due to copy&paste

stock_lot_name_sequence/wizard/stock_lot_suggest_name_wizard.py
```diff
@@ -0,0 +1,26 @@
...
+from odoo import api, fields, models
+
+
+class StockLotSuggestNameWizard(models.TransientModel):
+    _name = "stock.lot.suggest.name.wizard"
+    _description = "Suggest Lot/Serial Name"
+
+    lot_id = fields.Many2one("stock.lot", required=True, readonly=True)
+    product_id = fields.Many2one(related="lot_id.product_id", readonly=True)
+    current_name = fields.Char(related="lot_id.name", readonly=True)
```
dalonsod: Unneeded readonly attributes

stock_lot_name_sequence/wizard/stock_lot_suggest_name_wizard_views.xml
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+
+    <record id="view_stock_lot_suggest_name_wizard_form" model="ir.ui.view">
+        <field name="name">stock.lot.suggest.name.wizard.form</field>
+        <field name="model">stock.lot.suggest.name.wizard</field>
+        <field name="arch" type="xml">
+            <form>
+                <field name="lot_id" invisible="1"/>
```
dalonsod, line 9: Should this be removed, as unused at the view?

stock_lot_name_sequence/__manifest__.py
```diff
@@ -0,0 +1,20 @@
+# Copyright 2026 Solvos Consultoría Informática, S.L. (<https://www.solvos.es>)
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl).
+{
+    "name": "Stock Lot Name Sequence",
+    "version": "19.0.1.0.0",
+    "category": "Inventory",
+    "website": "https://github.com/solvosci/slv-manufacture",
```
dalonsod: Right now this is still here, but it seems to be better placed at `slv-stock`, what do you think?

dalonsod (changes requested):
Only code review, see comments.

I tried to test this, but I finally had no time, sorry....

mrp_production_classification/models/mrp_production.py
cgarcia-solvos: Needed for the view: the "Classify" button's `invisible` condition needs this field to show it only on orders whose operation type is marked "Is Classification Operation".

mrp_production_classification/models/mrp_production.py
cgarcia-solvos: That's intentional, Odoo aggregates all issues (e.g. consumption warnings) into one wizard per button_mark_done() call. For classification orders we want to skip that wizard; but with a mixed selection there's only one call, so splitting it would mean processing one group and silently leaving the other pending. Instead, mixed selections fall back to standard behavior (no skip), so every selected order still gets processed and none is deprived of its wizard.

stock_lot_name_sequence/wizard/stock_lot_suggest_name_wizard_views.xml
cgarcia-solvos: Can't remove it. `product_id` and `current_name` are both related to `lot_id`. Without lot_id in the view, they end up empty.

stock_lot_name_sequence/__manifest__.py
cgarcia-solvos: Yes. Better.

stock_lot_name_sequence/models/stock_lot.py
cgarcia-solvos: Can't believe I missed one....

